PR #4, opened 2021-02-03
:cop: Format code

ookami-kb:


chaeMil (approved):
I presume i dont really need to check all the 56 files :D

PR #8, opened 2021-02-03
:zap: [RND-43393] Develop Banner component

chaeMil:


optimus/lib/src/banner.dart
```diff
@@ -0,0 +1,261 @@
...
+            children: [
+              Padding(
+                padding: _padding,
+                child: Row(
+                  crossAxisAlignment: CrossAxisAlignment.center,
+                  children: [
+                    if (showIcon)
+                      Padding(
+                        padding: const EdgeInsets.only(right: 18),
+                        child: Icon(_icon, size: 20, color: _iconColor),
```
Gladkowski, line 67: I think that size 20 is not particularly correct. So I suggest replacing it with `OptimusIcon`, with medium size and relevant OptimusColorOption.

optimus/lib/src/banner.dart
```diff
@@ -0,0 +1,261 @@
...
+        return OptimusColors.danger500;
+    }
+  }
+
+  // ignore: missing_return
+  Color get _color {
+    switch (variant) {
+      case OptimusWideBannerVariant.informative:
+      case OptimusWideBannerVariant.danger:
+        return Colors.white;
```
Gladkowski, line 256: `return OptimusColors.basic0;`

optimus/lib/src/banner.dart
```diff
@@ -0,0 +1,261 @@
...
+      case OptimusWideBannerVariant.warning:
+        return Colors.black;
```
Gladkowski, line 258: `return OptimusColors.basic1000;`

optimus/lib/src/banner.dart
```diff
@@ -0,0 +1,261 @@
...
+                      ],
+                    ),
+                  ],
+                ),
+              ),
+              if (dismissible)
+                Positioned(
+                  right: 6,
+                  top: 10,
+                  child: GestureDetector(
```
Gladkowski, line 93: I would replace it with something like that:
```
 // TODO(): Replace with OptimusIconButton when ready
 // https://mews.myjetbrains.com/youtrack/issue/RND-47422
IconButton(
                splashRadius: 24,
                padding: EdgeInsets.zero,
                onPressed: () => Navigator.pop(context),
                icon: const OptimusIcon(iconData: OptimusIcons.cross_close),
              ),
```

ookami-kb:
@chaeMil 

- warning: `OptimusIcons.problematic`
- error: `OptimusIcons.blacklist`

optimus/lib/src/banner.dart
```diff
@@ -0,0 +1,261 @@
...
+
+  // ignore: missing_return
+  IconData get _icon {
+    switch (variant) {
+      case OptimusBannerVariant.primary:
+        return OptimusIcons.info;
+      case OptimusBannerVariant.success:
+        return OptimusIcons.done_circle;
+      case OptimusBannerVariant.warning:
+        return OptimusIcons.disable; // TODO(MM): add warning icon
```
Gladkowski, line 141: `return OptimusIcons.problematic;`

optimus/lib/src/banner.dart
```diff
@@ -0,0 +1,261 @@
...
+      case OptimusBannerVariant.error:
+        return OptimusIcons.disable; // TODO(MM): update icon
```
Gladkowski, line 143: `return OptimusIcons.blacklist;`

Gladkowski (changes requested):

optimus/lib/src/banner.dart
ookami-kb: `IconButton` is already merged, so it can be used here.

Gladkowski (approved):

optimus/lib/src/banner.dart
```diff
@@ -0,0 +1,267 @@
...
+    this.variant = OptimusBannerVariant.primary,
+    this.showIcon = false,
+    this.additionalDescription,
+    this.dismissible = false,
+    this.onDismiss,
+  }) : super(key: key);
+
+  final Widget content;
+  final OptimusBannerVariant variant;
+  final bool showIcon;
```
ookami-kb, line 45: Let's name it as `hasIcon` – booleans starting with `is*` or `has*` are better readable. `showIcon` sounds like a method name.

optimus/lib/src/banner.dart
```diff
@@ -0,0 +1,267 @@
...
+                ),
+            ],
+          ),
+        ),
+      );
+
+  bool get _hasAdditionalDescription =>
+      additionalDescription?.isNotEmpty ?? false;
+
+  double get _bannerHeight => _hasAdditionalDescription ? 68 : 40;
```
ookami-kb, line 112: I don't think we should hard-code the size.

optimus/lib/src/banner.dart
```diff
@@ -0,0 +1,267 @@
...
+    this.showIcon = false,
+    this.additionalDescription,
+    this.dismissible = false,
+    this.onDismiss,
+  }) : super(key: key);
+
+  final Widget content;
+  final OptimusBannerVariant variant;
+  final bool showIcon;
+  final String additionalDescription;
```
ookami-kb, line 46: I would rename `content` to `title` and `additionalDescription` to `description`. Also, `description` can be made `Widget` type.

optimus/lib/src/banner.dart
```diff
@@ -0,0 +1,267 @@
...
+
+  @override
+  Widget build(Object context) => ConstrainedBox(
+        constraints: BoxConstraints(maxHeight: _bannerHeight),
+        child: Container(
+          decoration: BoxDecoration(
+            color: _backgroundColor,
+            borderRadius: const BorderRadius.all(borderRadius50),
+          ),
+          child: Stack(
```
ookami-kb, line 58: Let's make it a row with an expanded title and close button.

optimus/lib/src/banner.dart
```diff
@@ -0,0 +1,267 @@
+import 'package:flutter/material.dart';
+import 'package:flutter/widgets.dart';
+import 'package:optimus/optimus.dart';
+import 'package:optimus/src/border_radius.dart';
+import 'package:optimus/src/typography/styles.dart';
+
+enum OptimusBannerVariant {
```
ookami-kb, line 7: Do we actually need it as a separate `enum`? It matches 1:1 with `OptimusColorOption` and I think it has the same semantic meaning.

ookami-kb (changes requested):

optimus/lib/src/banner.dart
chaeMil: `OptimusColorOption` has `basic` which is not present here. That why i created that enum.

optimus/lib/src/banner.dart
chaeMil: Im not sure how to properly place the close button then. The `OptimusIconButton` is too big and then it expands the whole `Banner` more than in the design, this is what i mean: 
image

ookami-kb: A title shouldn't be that close to the top border, there's padding there:
image


> it expands the whole Banner more than in the design

Do you mean the width? But this is expected behaviour:

> A banner always takes the full width of the component it is within.

chaeMil: No i mean height, the close button when used as `OptimusIconButton` is just too big. Or solution would be to have two `Rows` inside each other. The would have ten inner one with, icon, labels and then next to the inner row would be the close button without paddings so it will fill the space correctly.

ookami-kb:
@chaeMil @Gladkowski I've slightly updated formatting and added docs.

Also, I've removed `Link` from the system-wide banner. It's rather up for a user to set the content, otherwise, it will complicate the layout in case of long text content, etc. Plus, I guess the link would rather be a separate component in the future.

ookami-kb (approved):

PR #9, opened 2021-02-04
:cop: [RND-48205] Update Optimus colors

Gladkowski:


optimus/lib/src/tabs.dart
```diff
@@ -45,9 +45,9 @@ class OptimusTabBar extends StatelessWidget {
               child: TabBar(
                 controller: tabController,
                 indicator: _buildIndicator(),
-                unselectedLabelColor: OptimusColors.basic900t64,
-                labelColor: OptimusColors.basic900,
-                indicatorColor: OptimusColors.basic900,
+                unselectedLabelColor: OptimusColors.neutral1000t64,
```
ookami-kb, line 48: Why `basic900t64` is replaced with `neutral1000t64` (not with `neutral900t64`)? Here and in other places.

optimus/lib/src/colors/colors.dart
```diff
@@ -1,103 +1,154 @@
...
-  static const basic100 = Color(0xFFC0C5CD);
-  static const basic200 = Color(0xFF969FAB);
-  static const basic300 = Color(0xFF6B7989);
-  static const basic400 = Color(0xFF4C5C70);
-  static const basic500 = Color(0xFF2C3F57);
-  static const basic700 = Color(0xFF213146);
-  static const basic800 = Color(0xFF1B293C);
-  static const basic900 = Color(0xFF101B2C);
-  static const basic1000 = Color(0xFF000000);
+  /// Neutral colors
```
ookami-kb, line 13: A comment like this will refer only to `neutral0` which makes a little sense. Let's maybe move all these comments to the main `OptimusColors` comment?

ookami-kb (changes requested):

optimus/lib/src/tabs.dart
Gladkowski: Not sure why, but it was replaced for neutral color. (Checked with designers).

ookami-kb (approved):

PR #16, opened 2021-02-17
:cop: [RND-48571] Button animation

chaeMil:
#### Summary

Make animations consistent in `Button`, `IconButton` and `DropdownButton`

#### Testing steps

#### Follow-up issues

#### Check during review

- Unused code removed.
- Build passing.


optimus/lib/src/button.dart
```diff
@@ -170,6 +171,7 @@ class OptimusButton extends StatelessWidget {
           disabledColor: _color,
           disabledTextColor: _textColor,
           hoverElevation: 0,
+          splashColor: Colors.transparent,
```
Gladkowski, line 174: I would rather replace the `MaterialButton` with the same structure as we have in dropdown_button and icon_button. And then extract it and reuse across all the button components we have.

Gladkowski (changes requested):

optimus/lib/src/button.dart
chaeMil: Good idea but the problem is with the circular icon button because it have different box decoration and all the 3 implementations have the wrapping widgets setup little bit differently. Also Im not sure if the propagation of `isHovering` and `isTappedDown` would make things easier to work with. 

Any opinions guys? @ookami-kb @Gladkowski

ookami-kb: Let's probably leave it like this in the current iteration, and discuss it later.

optimus/lib/src/button/dropdown.dart
```diff
@@ -51,6 +51,7 @@ class _OptimusDropDownButtonState<T> extends State<OptimusDropDownButton<T>> {
   bool _isTappedDown = false;
   bool _isOpened = false;
   final _node = FocusNode();
+  static const _animationDuration = Duration(milliseconds: 150);
```
ookami-kb, line 54: Extract this constant to some common folder and reuse it in all places. Also, it should be explicitly set in `OptimusButton`.

ookami-kb (changes requested):

ookami-kb (approved):

PR #19, opened 2021-02-18
:cop: [RND-48793] Implement split button part 1

Gladkowski:
#### Summary

- add base class for button
- add base class for dropdown button


optimus/lib/src/button/button.dart
```diff
@@ -61,6 +61,51 @@ class OptimusButton extends StatelessWidget {
...
+    Key key,
+    this.onPressed,
+    @required this.child,
+    this.minWidth,
+    this.leftIcon,
+    this.rightIcon,
+    this.badgeLabel,
+    this.size = OptimusWidgetSize.large,
+    this.variant = OptimusButtonVariant.defaultButton,
+    this.borderRadius = const BorderRadius.all(borderRadius50),
```
ookami-kb, line 88: `borderRadius` should be provided by `OptimusButton`, as it will be different for split button. Same for dropdown button.

ookami-kb (changes requested):
Let's extract `Base*` components to separate files.

ookami-kb (approved):

PR #20, opened 2021-02-19
:zap: [RND-48793] Implement split button part 2

Gladkowski:


optimus/lib/src/button/button.dart
```diff
@@ -4,6 +4,7 @@ import 'package:optimus/optimus.dart';
 import 'package:optimus/src/button/base_button.dart';
 import 'package:optimus/src/widget_size.dart';
 
+// TODO(VG): rename to OptimusButtonType
```
ookami-kb, line 7: I would say it's vice versa – according to documentation `OptimusSplitButtonType` should be renamed to `OptimusSplitButtonVariant`.

Gladkowski: Ok, will replace it in the task too

ookami-kb (changes requested):
`size` is not working.

optimus/lib/src/button/split.dart
```diff
@@ -0,0 +1,91 @@
...
+/// Split buttons have a main action and a dropdown action. The main action is
+/// on the left. An arrow on the right opens a dropdown menu with more actions
+/// related to the main action.
+class OptimusSplitButton<T> extends StatelessWidget {
+  const OptimusSplitButton({
+    Key key,
+    @required this.child,
+    this.onPressed,
+    @required this.items,
+    this.onDropdownItemPressed,
```
ookami-kb, line 25: Since we named it `onItemSelected` in dropdown button, let's go with the same naming here.

ookami-kb (approved):

PR #31, opened 2021-02-25
:zap: [RND-49612] Add dark theme for typography

Gladkowski:


optimus/lib/src/colors/colors.dart
```diff
@@ -284,6 +284,12 @@ class OptimusColors {
 
   final Brightness brightness;
 
+  Color get defaultTextColor => OptimusDarkColors.neutral1000;
```
ookami-kb, line 287: Why `OptimusDarkColors`? Shouldn't it just be `neutral1000`? And same with the others.

ookami-kb (approved):

PR #38, opened 2021-03-01
:zap: [RND-49612] Add theming for badge

Gladkowski:
- modified badge to stop it from stretching to parent's size
- added badge story
- added theming for badge


ookami-kb:
@Gladkowski I've pushed some update: looks like text style is different from any presets so far. Also, I've further simplified layout.

Gladkowski:
> @Gladkowski I've pushed some update: looks like text style is different from any presets so far. Also, I've further simplified layout.

Nice additions, lgtm.

ookami-kb (approved):

PR #41, opened 2021-03-03
:zap: [RND-49612] Add dark theme colors for tabs and borders

Gladkowski:


ookami-kb:
@Gladkowski I've put some changes: I've made `borderSide` property private to the library – I don't think it should be used outside of it. Instead, I've added `borderColor` property to the theme.

Also, I've slightly simplified the story.

Gladkowski:
> @Gladkowski I've put some changes: I've made `borderSide` property private to the library – I don't think it should be used outside of it. Instead, I've added `borderColor` property to the theme.
> 
> Also, I've slightly simplified the story.

Nice. I've just updated tab content color to make text visible for both dark and light theme.

ookami-kb (approved):

PR #47, opened 2021-03-03
:zap: [RND-49612] Add dark theme support for search field

Gladkowski:


optimus/lib/src/search/search_field.dart
```diff
@@ -72,29 +72,35 @@ class _OptimusSearchState<T> extends State<OptimusSearch<T>> {
...
-                ),
-          isEnabled: widget.isEnabled,
-          caption: widget.caption,
-          secondaryCaption: widget.secondaryCaption,
-          error: widget.error,
-          size: widget.size,
-        ),
-      );
+  Widget build(BuildContext context) {
+    final theme = OptimusTheme.of(context);
```
ookami-kb, line 76: Since the theme is used by Icon only, it's better to extract `Icon` into a separate widget and avoid rebuilding the whole widget.

ookami-kb (changes requested):

ookami-kb (approved):

PR #48, opened 2021-03-03
:zap: [RND-49612] Add dark theme support for select

Gladkowski:


optimus/lib/src/select.dart
```diff
@@ -93,29 +93,35 @@ class _OptimusSelectState<T> extends State<OptimusSelect<T>> {
...
+          return preset200m.copyWith(color: _textColor);
         case OptimusWidgetSize.medium:
         case OptimusWidgetSize.large:
-          return preset300m.copyWith(color: OptimusLightColors.neutral900);
+          return preset300m.copyWith(color: _textColor);
       }
     }
   }
+
+  Color get _lightTextColor =>
```
ookami-kb, line 120: Why is it called `_lightTextColor`? Isn't it rather placeholder color?

ookami-kb (approved):

PR #52, opened 2021-03-05
:cop: Migrate to null-safety [ignore-size]

ookami-kb:
#### Summary

Migrated all the code to support null-safety. There are a lot of changes, but they are mainly trivial, so it doesn't make sense to break it into multiple PRs.

#### Testing steps

No.

#### Follow-up issues

No.

#### Check during review

- Unused code removed.
- Build passing.


optimus/lib/src/search/base_dropdown_tile.dart
```diff
@@ -26,7 +26,8 @@ class _BaseDropdownTileState extends State<BaseDropdownTile> {
           crossAxisAlignment: CrossAxisAlignment.start,
           children: <Widget>[
             OptimusHighlightLow(child: widget.title),
-            if (widget.subtitle != null) OptimusSubtitle(child: widget.subtitle)
+            if (widget.subtitle != null)
+              OptimusSubtitle(child: widget.subtitle!)
```
chaeMil, line 30: What about using `?.let` in cases like this?

ookami-kb: Won't work 😞 as it will fallback to `null`. With `if` there wouldn't be element at all.

Gladkowski (approved):

optimus/lib/src/search/base_dropdown_tile.dart
chaeMil: Ok, otherwise LGTM

chaeMil (approved):

PR #57, opened 2021-03-12
feat(optimus): Add showCursor parameter

ookami-kb:
#### Summary

Added parameter `showCursor` + some "free" documentation.

#### Testing steps

No.

#### Follow-up issues

No.

#### Check during review

- Unused code removed.
- Build passing.


optimus/lib/src/number_picker.dart
```diff
@@ -14,6 +14,8 @@ class OptimusNumberPicker extends StatefulWidget {
     this.focusNode,
     this.isEnabled = true,
     this.error,
+    this.readOnly = false,
+    this.showCursor,
```
Gladkowski, line 18: Should it have the default value?

Gladkowski (changes requested):

optimus/lib/src/number_picker.dart
ookami-kb: No, default value is `null` which means that behaviour depends on `readOnly` parameter.

Gladkowski (approved):

PR #58, opened 2021-03-19
feat(optimus): Make enabled public

Gladkowski:


chaeMil (approved):

ookami-kb (changes requested):
Since we're exporting it, let's at least rename it as `OptimusEnabled`.

ookami-kb (approved):

PR #60, opened 2021-03-22
feat(optimus): [RND-52584] Add segmented control 

Gladkowski:
#### Summary

@chaeMil @ookami-kb I'm experiencing issues with borders.

borders

The initial idea was to create a container with just 3 borders, but it turns out [it is not possible when radius is non-null](https://github.com/flutter/flutter/issues/12583).

I tried ClipRect but it leaves empty space instead of border that I am trying to wipe.

Any ideas on how to solve it without custom border drawing tool are welcome. Thanks ;)


Gladkowski:
@ookami-kb @chaeMil ready to review

chaeMil (approved):

optimus/lib/src/segmented_control/segmented_control.dart
```diff
@@ -0,0 +1,177 @@
...
+                    ),
+                  ),
+                ),
+              ],
+            ),
+          ),
+        ),
+      );
+
+  Color get _textColor => _isSelected
```
ookami-kb, line 168: According to design, `textColor` for light theme is always `neutral900` for both selected and non-selected.

optimus/lib/src/segmented_control/segmented_control.dart
```diff
@@ -0,0 +1,177 @@
...
+    this.isRequired = false,
+  }) : super(key: key);
+
+  final OptimusWidgetSize size;
+
+  final List<OptimusGroupItem<T>> items;
+
+  final T value;
+
+  final ValueChanged<T> onChanged;
```
ookami-kb, line 31: `onItemSelected` as for select control?

optimus/lib/src/segmented_control/segmented_control.dart
```diff
@@ -0,0 +1,177 @@
...
+  final bool isRequired;
+
+  @override
+  _OptimusSegmentedControlState<T> createState() =>
+      _OptimusSegmentedControlState<T>();
+}
+
+class _OptimusSegmentedControlState<T>
+    extends State<OptimusSegmentedControl<T>> {
+  int _selectedItemIndex = 0;
```
ookami-kb, line 48: The component shouldn't maintain its state. The selected value comes from the client.

optimus/lib/src/segmented_control/wrapper.dart
```diff
@@ -0,0 +1,136 @@
...
+              ),
+            ),
+          ],
+        ),
+      );
+
+  List<Widget> _dividers(double maxWidth) =>
+      Iterable<int>.generate(widget.listSize - 1)
+          .map((e) => IgnorePointer(
+                ignoring: true,
```
ookami-kb, line 65: It's `true` by default.

optimus/lib/src/segmented_control/segmented_control.dart
```diff
@@ -0,0 +1,177 @@
...
+import 'package:optimus/src/group_wrapper.dart';
+import 'package:optimus/src/segmented_control/wrapper.dart';
+import 'package:optimus/src/typography/styles.dart';
+
+class OptimusSegmentedControl<T> extends StatefulWidget {
+  const OptimusSegmentedControl({
+    Key? key,
+    this.size = OptimusWidgetSize.large,
+    required this.items,
+    required this.value,
```
ookami-kb, line 17: I would limit the value. I guess segmented control should *always* have some value out of provided, so that one value is always "on".

So maybe an assert that checks that value is in provided items?

ookami-kb (changes requested):

ookami-kb:
@Gladkowski @chaeMil I've pushed some small updates, please check.

Gladkowski:
> @Gladkowski @chaeMil I've pushed some small updates, please check.

Checked. Looks good.

ookami-kb (approved):

PR #63, opened 2021-03-29
fix(optimus): Fix layout issues with dispose

ookami-kb:
#### Summary

It happens if post frame callback happens after component is disposed. E.g. with long lists after scrolling back and forth.

#### Testing steps

Open `Slidable` story and scroll back and forth quickly. No errors should happen.

#### Follow-up issues

No.

#### Check during review

- Unused code removed.
- Build passing.


chaeMil (approved):
Noice! :)

PR #73, opened 2021-05-03
fix(optimus): Add `isRequired` to multiple field components

Sevastyan:
#### Summary

1. Added `isRequired` support to `OptimusCheckbox`. Updated the story.

2. Added `isRequired` support to `OptimusCheckboxGroup`. Updated the story.

3. Added `isRequired` support to `OptimusDateTimeField`.

4. Re-written `OptimusCheckboxFormField`, because it was not efficiently
implemented. 

5. Added `isRequired` support to `OptimusSelectFormField`.

6. Updated `form` story.

I've also moved `isRequired` next to `label` property in some places. This is
because `isRequired` affects `label`, and it is more convenient to have them
together because of it.

#### Testing steps

No.

#### Follow-up issues

No.

#### Check during review

- Unused code removed.
- Build passing.


Gladkowski (approved):

optimus/lib/src/checkbox/checkbox.dart
```diff
@@ -37,7 +39,12 @@ class OptimusCheckbox extends StatefulWidget {
   /// Label displayed next to checkbox.
   ///
   /// Typically a [Text] widget.
-  final Widget label;
+  final String label;
```
ookami-kb, line 42: We cannot change it to `String`. We're already using it as a widget, e.g. for displaying terms and conditions as text with links.

optimus/lib/src/group_item.dart
```diff
@@ -8,6 +6,6 @@ class OptimusGroupItem<T> {
     required this.value,
   });
 
-  final Widget label;
+  final String label;
```
ookami-kb, line 9: Same here, it can be useful to provide some custom widget, and we're already doing it by providing `TranslatedText`.

ookami-kb (changes requested):

optimus/lib/src/checkbox/checkbox.dart
Gladkowski: Oops, forgot about that. Good point

chaeMil (approved):

Sevastyan:
Closing as the solution is not ideal:

* asterisk is not looking good for multi-line text .
* it may be even worth, as the label is `Widget`, which can be anything.

Clients can add asterisk themselves, if they so desire.

PR #77, opened 2021-05-11
feat: [RND-58015] Add kiosk_mode package

Gladkowski:
#### Summary

Add kiosk mode support for Android. You can check sample app under the example directory.

Only basic functionality is present, and there are couple of open questions.
 - should the `hide interface` be controlled by the package itself or by the app? (now it is controlled by the package)
 - how should we react when user declines to pin the screen?
 - do we need some method that says us whether the screen is pinned right now?

According to documentation - _Don't start lock task mode when the device is locked because the user might not be able to unlock the device. You can call KeyguardManager methods to find out if the device is locked and use an Activity lifecycle callback (such as onResume() that's called after unlocking) to start lock task mode._ - not sure if we need to implement it.

All suggestions are welcome.


chaeMil:
> should the hide interface be controlled by the package itself or by the app? (now it is controlled by the package)
need to be controlled by the app it can change depending on the app mode (`kiosk` or `notKiosk`)

> how should we react when user declines to pin the screen?
user cannot decline this, the app has all permission granted through DPC

> do we need some method that says us whether the screen is pinned right now?
i guess no, we switch it with the `kiosk` / `notKiosk` mode in flutter and thats it

ookami-kb:
@Gladkowski I've pushed some updates, please check.

Gladkowski:
> @Gladkowski I've pushed some updates, please check.

Looks good, thanks for adding the test

PR #79, opened 2021-05-13
fix(optimus): Fix error on form fields

Sevastyan:
#### Summary

When rebuilding a form field widget, some clients need to update `error` text on the form field. The only way to update it, is to call `validator`.

Use-case in Mews: we rebuild form field widgets on a localization context change. With the change of localization context, we expect error text to be updated with a new translation.

Downside of the solution this PR suggests is that for those clients, that don't need to update `error` during form field widget rebuild, are going to have their `validator` called. We can introduce a constructor parameter for form field widgets to configure this behaviour. What do you think?

#### Testing steps

No.

#### Follow-up issues

No.

#### Check during review

- Unused code removed.
- Build passing.

### DIFF
--- a/optimus/lib/src/form/date_time_form_field.dart
+++ b/optimus/lib/src/form/date_time_form_field.dart
@@ -24,7 +24,7 @@ class OptimusDateTimeFormField extends FormField<DateTime> {
             onChanged: (v) => field.didChange(v),
             minDate: minDate,
             maxDate: maxDate,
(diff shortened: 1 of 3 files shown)


ookami-kb:
It's not for the library to decide whether the validator should be called again.

Also, it's a bad idea to call some side-effect method during the build, it can lead to some subtle errors. Regarding the use-case with translations, we can have a separate discussion.

PR #85, opened 2021-06-17
chore(release): publish packages

Gladkowski:
#### Summary

 - optimus@0.6.1


ookami-kb:
@Gladkowski how comes that breaking change introduces patch version update? Did you use automatic version generation?

PR #89, opened 2021-06-22
fix: Fix dropdown text color

Sevastyan:


Gladkowski (approved):
I like this solution. It should be reused all across the lib, replacing `DefaultTextStyle.merge` where possible.

ookami-kb (changes requested):
It breaks the selected text color in light theme mode (the text should be white):

Screenshot 2021-06-22 at 11 11 01

Sevastyan:
@ookami-kb This PR doesn't break the color of the highlighted text, because it has been broken from the beginning. This PR just "highlights" the problem (pun intended).

I think that this PR is worth merging. It objectively makes the situation better, while complying with the presets. Someone can later fix the highlight, cause I don't see an easy solution at the moment.

drop

ookami-kb:
No, it makes the situation worse in the library itself. The screenshot I provided is from optimus storybook. Previously, in the light theme, the highlighted list item had a white color, now it's black. We can discuss possible solution in a call.

Sevastyan:
The pictures I've provided are legit. 

>  in the light theme, the highlighted list item had a white color

is the picture number 2 from the left column (counting starts with 1). So, it **was** (and still is) broken.

ookami-kb:
Take a look at the optimus storybook "Select" page before and after.

Sevastyan:
It is broken in the storybook before:

before

ookami-kb:
Yes, but only subtitle was broken, now the title is broken as well.

Sevastyan:
Well, yes, but the overall it got better, because it is not broken in the dark theme.

ookami-kb:
I cannot call repairing a dark theme by breaking the light theme a "better" overall situation.

Sevastyan:
@ookami-kb I've updated the solution.

ookami-kb:
I don't like that it couples things together and introduces an obscure parameter on `OptimusTypography`. I would rather propose changing `_DropdownItemState`:

```dart
        child: _isHighlighted
            ? OptimusTheme(
                themeMode: ThemeMode.dark,
                darkTheme: OptimusTheme.of(context)
                    .copyWith(brightness: Brightness.dark),
                child: widget.child,
              )
            : widget.child,
```

Basically, we say that highlighted tile has a dark theme – which is true, it's light text on dark background.

ookami-kb:
Please remove `kiosk_mode` files from PR, otherwise LGTM.

ookami-kb (approved):

PR #92, opened 2021-06-24
fix(optimus): [RND-63238] Fix long dialog title

Gladkowski:
#### Summary

Dialog title is multiline now

### DIFF
--- a/optimus/lib/src/dialogs/dialog.dart
+++ b/optimus/lib/src/dialogs/dialog.dart
@@ -363,18 +363,28 @@ class _Title extends StatelessWidget {
   final bool isDismissible;
 
   @override
-  Widget build(BuildContext context) => Row(
+  Widget build(BuildContext context) => Stack(
         children: [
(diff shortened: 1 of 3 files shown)


Gladkowski:
@chaeMil ping

chaeMil (approved):

Gladkowski:
@ookami-kb ping

optimus/lib/src/dialogs/dialog.dart
```diff
@@ -363,18 +363,28 @@ class _Title extends StatelessWidget {
   final bool isDismissible;
 
   @override
-  Widget build(BuildContext context) => Row(
+  Widget build(BuildContext context) => Stack(
```
ookami-kb, line 366: Why switching from `Row` to `Stack`? It makes the layout more fragile, as you need to take right padding into account.

ookami-kb (approved):

PR #93, opened 2021-06-24
fix: Fix OptimusSelect focus

ookami-kb:
#### Summary

Added `Focus` wrapper in the `OptimusSelect`. 

The problem was that `FocusNode` itslef cannot receive focus, it should be passed to some focusable widget. `OptimusInputField` passes this `focusNode` to the underlying text field, so this is ok, but `OptimusSelect` doesn't have such focusable widget.

This focus node is also passed to field wrapper, bu this is ok, since it only listens to this node.

#### Testing steps

Selects should work.

#### Follow-up issues

No.

#### Check during review

- Unused code removed.
- Build passing.


Sevastyan (approved):

Sevastyan:
> FocusNode itslef cannot receive focus, it should be passed to some focusable widget

🤦🏻

PR #95, opened 2021-06-29
feat(optimus): [RND-63261] Optimus Chat Bubbles [ignore-size]

chaeMil:
#### Summary

Added `OptimusChatBubble`, `OptimusChatInput` and `OptimusChat` Widgets.

#### Testing steps

#### Follow-up issues

#### Check during review

- Unused code removed.
- Build passing.


storybook/lib/stories/chat/bubble.dart
```diff
@@ -0,0 +1,80 @@
...
+
+      return Center(
+        child: OptimusChatBubble(
+          index: 0,
+          message: message,
+          showAvatar: k.boolean(label: 'Show avatar', initial: true),
+          showUserName: k.boolean(label: 'Show user name', initial: true),
+          showStatus: k.boolean(label: 'Show status', initial: true),
+          showDate: k.boolean(label: 'Show date', initial: true),
+          formatTime: (DateTime input) => '${input.hour}:${input.minute}',
```
Gladkowski, line 49: If the time is 14:05 time formats like `14:5`. I suggest using `DateFormat('hh:mm')`.

optimus/lib/src/chat/bubble.dart
```diff
@@ -0,0 +1,274 @@
...
+
+  Padding _status(OptimusThemeData theme) => Padding(
+        padding: _statusPadding,
+        child: DefaultTextStyle.merge(
+          style: TextStyle.lerp(
+            preset100s,
+            TextStyle(
+                color: theme.isDark
+                    ? theme.colors.neutral0t64
+                    : theme.colors.neutral1000t64),
```
Gladkowski, line 127: Trailing comma

optimus/lib/src/chat/bubble.dart
```diff
@@ -0,0 +1,274 @@
...
+import 'package:flutter/widgets.dart';
+import 'package:freezed_annotation/freezed_annotation.dart';
+import 'package:optimus/optimus.dart';
+import 'package:optimus/src/border_radius.dart';
+import 'package:optimus/src/stack.dart';
+import 'package:optimus/src/typography/presets.dart';
+
+part 'bubble.freezed.dart';
+
+class OptimusChatBubble extends StatelessWidget {
```
Gladkowski, line 11: I think that comments should be added to all new classes and fields.

optimus/lib/src/chat/bubble.dart
```diff
@@ -0,0 +1,274 @@
...
+      case MessageStatus.notSent:
+        children = [
+          notSend,
+          DefaultTextStyle.merge(
+            style: const TextStyle(
+              decoration: TextDecoration.underline,
+            ),
+            child: GestureDetector(
+                onTap: () => onTryAgainClicked(index, message),
+                child: tryAgain),
```
Gladkowski, line 183: trailing comma

Also `tryAgain` should be wrapped in proper text style or typography component

optimus/lib/src/chat/bubble.dart
```diff
@@ -0,0 +1,274 @@
...
+            _messageBubble(theme),
+            if (message.alignment == MessageAlignment.right) _avatar,
+          ],
+        ),
+        if (showStatus) _status(theme),
+      ],
+    );
+  }
+
+  Widget _messageBubble(OptimusThemeData theme) => Flexible(
```
Gladkowski, line 68: Should be renamed to `_buildMessageBubble` since it's not a getter. Applies to all other places

optimus/lib/src/chat/bubble.dart
```diff
@@ -0,0 +1,274 @@
...
+  /// An error occurred. The message has not been sent.
+  notSent,
+}
+
+typedef TrySendAgain = Future<MessageStatus> Function(
+    int index, OptimusMessage message);
+
+@freezed
+class OptimusMessage with _$OptimusMessage {
+  const factory OptimusMessage({
```
Gladkowski, line 265: Also I feel like we don't need separate `OptimusMessage` class. All those fields could be part of bubble. But not sure about that

optimus/lib/src/chat/bubble.dart
```diff
@@ -0,0 +1,274 @@
...
+      case MessageColor.neutral:
+      case MessageColor.light:
+        return theme.colors.neutral1000;
+      case MessageColor.dark:
+        return theme.colors.neutral0;
+    }
+  }
+}
+
+enum MessageAlignment {
```
Gladkowski, line 238: I would place all enums at the top of the file and add comments

optimus/lib/src/chat/bubble.dart
```diff
@@ -0,0 +1,274 @@
...
+
+  Padding get _userName => Padding(
+        padding: _statusPadding,
+        child: DefaultTextStyle.merge(
+          style: preset100s,
+          child: Text(message.userName),
+        ),
+      );
+
+  Padding _status(OptimusThemeData theme) => Padding(
```
Gladkowski, line 119: It feels like `_status` can be extracted to separate class/file.

Gladkowski (changes requested):
- when status is `Sent` no check icon is shown in bubble story
- messages are not centered vertically in bubbles
- in dark mode the text color is black for light blue message bubble, is it supposed to be like that?

storybook/lib/stories/chat/bubble.dart
chaeMil: We dont have`intl` library in the optimus library, this is just basic example, you will provide your own function.

optimus/lib/src/chat/bubble.dart
chaeMil: This way you just provide `List<OptimusMessage>` to the `OptimusChat` widget. No need to creating all the widgets by yourself.

optimus/lib/src/chat/bubble.dart
chaeMil: Why?

chaeMil:
@Gladkowski ping

Gladkowski (approved):

optimus/lib/src/chat/bubble.dart
```diff
@@ -0,0 +1,300 @@
...
+import 'package:optimus/optimus.dart';
+import 'package:optimus/src/border_radius.dart';
+import 'package:optimus/src/stack.dart';
+import 'package:optimus/src/typography/presets.dart';
+
+part 'bubble.freezed.dart';
+
+/// Chat bubble component offers single configurable bubble. The main use of
+/// this widget is in OptimusChat widget but it can be used also stand-alone.
+
```
ookami-kb, line 13: There shouldn't be new line here.

optimus/lib/src/chat/bubble.dart
```diff
@@ -0,0 +1,300 @@
...
+  final Function formatTime;
+  final Function formatDate;
+  final Widget sending;
+  final Widget notSend;
+  final Widget sent;
+  final Widget tryAgain;
+  final TrySendAgain onTryAgainClicked;
+
+  @override
+  Widget build(BuildContext context) {
```
ookami-kb, line 47: `build` should be the last method.

optimus/lib/src/chat/bubble.dart
```diff
@@ -0,0 +1,300 @@
...
+  right,
+}
+
+enum MessageColor {
+  neutral,
+  light,
+  dark,
+}
+
+enum MessageStatus {
```
ookami-kb, line 275: Let's rather replace it with something like that:

```dart
@freezed
class MessageState with _$MessageState {
  const factory MessageState.sending({required String text}) =
      MessageStateSending;

  const factory MessageState.sent({required String text}) = MessageStateSent;

  const factory MessageState.error({
    required String text,
    required void Function() onTryAgain,
  }) = MessageStateError;
}
```

optimus/lib/src/chat/bubble.dart
```diff
@@ -0,0 +1,300 @@
...
+
+part 'bubble.freezed.dart';
+
+/// Chat bubble component offers single configurable bubble. The main use of
+/// this widget is in OptimusChat widget but it can be used also stand-alone.
+
+class OptimusChatBubble extends StatelessWidget {
+  const OptimusChatBubble({
+    Key? key,
+    required this.index,
```
ookami-kb, line 17: This widget should know nothing about its index.

optimus/lib/src/chat/bubble.dart
```diff
@@ -0,0 +1,300 @@
...
+    required this.message,
+    required this.showAvatar,
+    required this.showStatus,
+    required this.showUserName,
+    required this.showDate,
```
ookami-kb, line 22: According to our style, it should rather be something like `isDateVisible`, same for other parameters.

storybook/lib/stories/chat/bubble.dart
```diff
@@ -0,0 +1,81 @@
...
+final List<Option<MessageAlignment>> _type =
+    MessageAlignment.values.map((e) => Option(describeEnum(e), e)).toList();
+
+final List<Option<MessageColor>> _color =
+    MessageColor.values.map((e) => Option(describeEnum(e), e)).toList();
+
+final List<Option<Widget>> _avatar = [
+  const Option('No image', avatar2),
+  const Option('With image', avatar1),
+  Option('Organisation avatar', organisationAvatar),
```
ookami-kb, line 80: This is confusing, I would remove this option from bubble story at all.

optimus/lib/src/chat/bubble.dart
ookami-kb: Let's rather extract all the enums and models to a separate file `message.dart`.

ookami-kb (changes requested):
I've added all the comments separately by accident 🤦

optimus/lib/src/chat/bubble.dart
chaeMil: Why including the text? Its already in the `OptimusMessage`

ookami-kb: I mean text of the status here, i.e. "Sending", "Sent", and remove it from the widget constructor.

chaeMil:
@ookami-kb ping

optimus/lib/src/chat/bubble.dart
ookami-kb: This was not addressed.

optimus/lib/src/chat/bubble.dart
```diff
@@ -0,0 +1,294 @@
...
+    required String text,
+  }) = MessageStateSending;
+
+  const factory MessageState.sent({
+    required String text,
+  }) = MessageStateSent;
+
+  const factory MessageState.error({
+    required String text,
+    required void Function() onTryAgain,
```
ookami-kb, line 277: `void Function()` can be replaced with `VoidCallback`.

optimus/lib/src/chat/chat.dart
```diff
@@ -0,0 +1,97 @@
...
+    required this.formatTime,
+    required this.formatDate,
+    required this.tryAgain,
+    required this.onTryAgainClicked,
+    required this.onSendClicked,
+  }) : super(key: key);
+
+  final List<OptimusMessage> messages;
+  final Function formatTime;
+  final Function formatDate;
```
ookami-kb, line 23: `formatDate` and `formatDate` should have proper type, not just `Function`.

ookami-kb (changes requested):

optimus/lib/src/chat/chat.dart
```diff
@@ -0,0 +1,100 @@
...
+    required this.tryAgain,
+    required this.onTryAgainClicked,
+    required this.onSendClicked,
+  }) : super(key: key);
+
+  final List<OptimusMessage> messages;
+  final FormatTime formatTime;
+  final FormatDate formatDate;
+  final Widget tryAgain;
+  final TryAgainCallback onTryAgainClicked;
```
ookami-kb, line 25: `onTryAgainPressed`

optimus/lib/src/chat/chat.dart
```diff
@@ -0,0 +1,100 @@
...
+  final SendCallback onSendClicked;
```
ookami-kb, line 26: `onSendPressed`

optimus/lib/src/chat/chat.dart
```diff
@@ -0,0 +1,100 @@
...
+    Key? key,
+    required this.messages,
+    required this.formatTime,
+    required this.formatDate,
+    required this.tryAgain,
+    required this.onTryAgainClicked,
+    required this.onSendClicked,
+  }) : super(key: key);
+
+  final List<OptimusMessage> messages;
```
ookami-kb, line 21: For lists that a part of public API it's better to do defensive copying. Otherwise, it can be broken by client if the list is changed directly.

optimus/lib/src/chat/bubble.dart
```diff
@@ -0,0 +1,250 @@
...
+    Key? key,
+    required this.message,
+    required this.isAvatarVisible,
+    required this.isStatusVisible,
+    required this.isUserNameVisible,
+    required this.isDateVisible,
+    required this.formatTime,
+    required this.formatDate,
+    required this.tryAgain,
+    required this.onTryAgainClicked,
```
ookami-kb, line 22: `onTryAgainPressed`

optimus/lib/src/chat/bubble.dart
```diff
@@ -0,0 +1,250 @@
...
+  }
+
+  OptimusStackAlignment get _bubbleAlignment =>
+      message.alignment == MessageAlignment.left
+          ? OptimusStackAlignment.start
+          : OptimusStackAlignment.end;
+
+  Widget get _avatar => isAvatarVisible ? message.avatar : _emptyAvatarSpace;
+
+  Widget get _emptyAvatarSpace => const SizedBox(width: 40);
```
ookami-kb, line 196: Why is it hard-coded to `40`? What if we will pass avatars of different size? In that case bubbles without avatar would have incorrect padding.

optimus/lib/src/chat/input.dart
```diff
@@ -0,0 +1,52 @@
...
+    return OptimusInputField(
+      controller: _controller,
+      maxLines: 4,
+      minLines: 1,
+      suffix: Opacity(
+        opacity: _controller.value.text.isNotEmpty
+            ? OpacityValue.enabled
+            : OpacityValue.disabled,
+        child: GestureDetector(
+          onTap: () {
```
ookami-kb, line 32: It's called even if the text is empty. Also, you can reuse `OptimusEnabled` component instead.

optimus/lib/src/chat/input.dart
```diff
@@ -0,0 +1,52 @@
...
+
+    return OptimusInputField(
+      controller: _controller,
+      maxLines: 4,
+      minLines: 1,
+      suffix: Opacity(
+        opacity: _controller.value.text.isNotEmpty
+            ? OpacityValue.enabled
+            : OpacityValue.disabled,
+        child: GestureDetector(
```
ookami-kb, line 31: There's no visual feedback on button pressed.

ookami-kb (changes requested):

optimus/lib/src/chat/bubble.dart
chaeMil: Solved it kinda hacky way, lets see what do you think about it. I couldnt find a way to measure the widget without using `GlobalKey` so its what it is.

ookami-kb: Let's maybe at least move this `SizedBox` up the tree, so that existing avatars are wrapped as well.

optimus/lib/src/chat/input.dart
```diff
@@ -0,0 +1,61 @@
...
+      controller: _controller,
+      maxLines: 4,
+      minLines: 1,
+      suffix: OptimusEnabled(
+        isEnabled: _isSendEnabled,
+        child: GestureDetector(
+          onTapDown: (_) => setState(() => _isTappedDown = true),
+          onTapUp: (_) => setState(() => _isTappedDown = false),
+          onTapCancel: () => setState(() => _isTappedDown = false),
+          onTap: _isSendEnabled
```
ookami-kb, line 36: `OptimusEnabled` takes care of this. No need to explicitly disable tap.

optimus/lib/src/chat/input.dart
```diff
@@ -0,0 +1,61 @@
...
+              ? () {
+                  widget.onSendPressed(_controller.text);
+                  _controller.clear();
+                }
+              : null,
+          child: Transform.scale(
+            scale: _isTappedDown ? 0.7 : 1,
```
ookami-kb, line 43: Is it somehow aligned with designers?

optimus/lib/src/chat/bubble.dart
```diff
@@ -0,0 +1,251 @@
...
+          constraints: const BoxConstraints(maxWidth: 480),
+          decoration: _buildMessageBackground(theme),
+          padding: const EdgeInsets.only(
+            left: spacing100,
+            right: spacing100,
+            top: spacing50,
+            bottom: spacing100,
+          ),
+          child: DefaultTextStyle.merge(
+            style: TextStyle.lerp(preset200s,
```
ookami-kb, line 52: What's the point of `lerp` here? Why not `copyWith`?

optimus/lib/src/chat/bubble.dart
```diff
@@ -0,0 +1,251 @@
...
+        right: message.alignment != MessageAlignment.left ? 48 : 0,
+      );
+
+  Widget _buildStatusTextStyle(OptimusThemeData theme, Widget child) =>
+      DefaultTextStyle.merge(
+        style: baseTextStyle.copyWith(
+          color: theme.isDark
+              ? theme.colors.neutral0t64
+              : theme.colors.neutral1000t64,
+          fontSize: 12,
```
ookami-kb, line 107: Why is here some hard-coded text size? Why is it not aligned with text presets?

ookami-kb (changes requested):

optimus/lib/src/chat/chat.dart
```diff
@@ -0,0 +1,100 @@
...
+                tryAgain: tryAgain,
+                onTryAgainClicked: onTryAgainClicked,
+              ),
+            ),
+          ),
+          OptimusChatInput(onSendPressed: onSendClicked),
+        ],
+      );
+
+  List<OptimusMessage> get _messages => messages.reversed.toList();
```
ookami-kb: It doesn't make sense to keep the initial parameters and have getter to create a new list every time. Create it during class instantiation.

optimus/lib/src/chat/input.dart
chaeMil: Ehm, nope...

ookami-kb: It should be.

chaeMil: Confirmed, no visual feedback on button, the visual feedback is that message is sent.

optimus/lib/src/chat/bubble.dart
chaeMil: Its like in the design 🤷‍♂️ 
image

optimus/lib/src/chat/input.dart
ookami-kb: Weird, but ok.

optimus/lib/src/chat/bubble.dart
ookami-kb: Ok, let's leave it like this for now, but please raise this discussion with designers. AFAIK, there should be no hard-coded typography styles.

optimus/lib/src/chat/chat.dart
```diff
@@ -0,0 +1,100 @@
...
+  final TryAgainCallback onTryAgainPressed;
+  final SendCallback onSendPressed;
+
+  @override
+  Widget build(BuildContext context) => OptimusStack(
+        children: [
+          Expanded(
+            child: ListView.builder(
+              itemCount: _messages.length,
+              reverse: true,
```
ookami-kb, line 36: Why do you first reverse incoming messages and then build reversed list again?

optimus/lib/src/chat/input.dart
```diff
@@ -0,0 +1,51 @@
...
+}
+
+class _OptimusChatInputState extends State<OptimusChatInput> {
+  final _controller = TextEditingController();
+
+  bool get _isSendEnabled => _controller.value.text.isNotEmpty;
+
+  @override
+  Widget build(BuildContext context) {
+    _controller.addListener(() => setState(() {}));
```
ookami-kb, line 22: Don't add a listener in build method: with every build, a new listener will be created.

optimus/lib/src/chat/bubble.dart
```diff
@@ -0,0 +1,250 @@
...
+          decoration: _buildMessageBackground(theme),
+          padding: const EdgeInsets.only(
+            left: spacing100,
+            right: spacing100,
+            top: spacing50,
+            bottom: spacing100,
+          ),
+          child: DefaultTextStyle.merge(
+            style: preset200s.copyWith(color: _createMessageTextColor(theme)),
+            child: Text(message.message),
```
ookami-kb, line 53: You can pass `style` directly to `Text`, why wrap it into `DefaultTextStyle`?

ookami-kb (changes requested):

optimus/lib/src/chat/chat.dart
chaeMil: The `parameter` in the `ListBuilder` is used to automatically scroll the list to bottom and to begin from bottom. List of messages is reversed so the newly added message is on the bottom and also the logic is the same as they appears on the screen.

ookami-kb (approved):

PR #96, opened 2021-06-30
fix(optimus): [RND-63908] Fix text colors

Gladkowski:
#### Summary

- fix radio title color
- fix dialog 'X' button color
- refactor `default` and `secondary` colors 

#### Testing steps

#### Follow-up issues

#### Check during review

- Unused code removed.
- Build passing.

### DIFF
--- a/optimus/lib/src/colors/colors.dart
+++ b/optimus/lib/src/colors/colors.dart
@@ -284,6 +284,12 @@ class OptimusColors {
 
   final Brightness brightness;
 
+  Color get themeSensitiveDefaultTextColor =>
+      _isLight ? defaultTextColor : invertedTextColor;
+
+  Color get themeSensitiveSecondaryTextColor =>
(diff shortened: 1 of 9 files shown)


chaeMil (approved):

optimus/lib/src/colors/colors.dart
```diff
@@ -284,6 +284,12 @@ class OptimusColors {
 
   final Brightness brightness;
 
+  Color get themeSensitiveDefaultTextColor =>
```
ookami-kb, line 287: I think we did it wrong initially. Text colors should be theme specific without explicit mentioning, i.e.:

```dart
Color get defaultTextColor => _isLight ? neutral1000 : neutral0;
Color get invertedTextColor => _isLight ? neutral0 : neutral1000;
```

Same for the secondary color. But we need to check all the usages not to break them.

ookami-kb (changes requested):

ookami-kb (approved):

PR #105, opened 2021-07-15
feat(optimus): [RND-65933] Add standalone link

Gladkowski:
#### Summary

- add standalone link
- `inherit` and `external` properties are not implemented

#### Testing steps

#### Follow-up issues

#### Check during review

- Unused code removed.
- Build passing.


optimus/lib/src/standalone_link.dart
```diff
@@ -0,0 +1,88 @@
...
+          onEnter: (_) => _onHoverChanged(true),
+          onExit: (_) => _onHoverChanged(false),
+          child: GestureDetector(
+            onTap: widget.onPressed,
+            onTapDown: (_) => setState(() => _isTappedDown = true),
+            onTapUp: (_) => setState(() => _isTappedDown = false),
+            onTapCancel: () => setState(() => _isTappedDown = false),
+            child: Text.rich(
+              TextSpan(
+                children: [
```
chaeMil, line 64: Why is this here? It looks and function without it.

chaeMil (changes requested):

chaeMil (approved):

optimus/lib/src/standalone_link.dart
```diff
@@ -28,11 +28,13 @@ class OptimusStandaloneLink extends StatefulWidget {
     this.onPressed,
     required this.text,
     this.size = OptimusLinkSize.normal,
+    this.isInherit = false,
   }) : super(key: key);
 
   final VoidCallback? onPressed;
   final String text;
   final OptimusLinkSize size;
+  final bool isInherit;
```
ookami-kb, line 37: When using like this, it's even more confusing than before. How should user understand that `isInherit` makes the link color black?

Can we somehow make it aligned with optimus guidelines and create a more meaningful parameter name?

Gladkowski: What about `isDefault`?

optimus/lib/src/standalone_link.dart
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  @override
+  Widget build(BuildContext context) => OptimusEnabled(
+        isEnabled: widget.onPressed != null,
+        child: MouseRegion(
+          onEnter: (_) => _onHoverChanged(true),
+          onExit: (_) => _onHoverChanged(false),
+          child: GestureDetector(
+            onTap: widget.onPressed,
```
ookami-kb, line 70: Why is `onTap` used both here and in `_gestureRecognizer`?

optimus/lib/src/standalone_link.dart
```diff
@@ -0,0 +1,93 @@
...
+            onTapDown: (_) => setState(() => _isTappedDown = true),
+            onTapUp: (_) => setState(() => _isTappedDown = false),
+            onTapCancel: () => setState(() => _isTappedDown = false),
+            child: Text.rich(
```
ookami-kb, line 72: Actually, I think we don't need rich text at all for a standalone link.

Gladkowski: well, valid idea

ookami-kb (approved):

PR #106, opened 2021-07-15
feat(optimus): [RND-65864] Change dialog button color

Gladkowski:
#### Summary

Change dialog's single button color to primary

#### Testing steps

#### Follow-up issues

#### Check during review

- Unused code removed.
- Build passing.

### DIFF
--- a/optimus/lib/src/dialogs/dialog.dart
+++ b/optimus/lib/src/dialogs/dialog.dart
@@ -438,7 +438,6 @@ class _Actions extends StatelessWidget {
   }
 
   OptimusButtonVariant _getVariant(int i) {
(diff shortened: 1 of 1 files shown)


optimus/lib/src/dialogs/dialog.dart
```diff
@@ -438,7 +438,7 @@ class _Actions extends StatelessWidget {
   }
 
   OptimusButtonVariant _getVariant(int i) {
-    if (actions.length == 1) return OptimusButtonVariant.defaultButton;
+    if (actions.length == 1) return OptimusButtonVariant.primary;
```
ookami-kb, line 441: Not needed at all, it's covered by line `445`.

ookami-kb (approved):

PR #112, opened 2021-07-21
fix(kiosk_mode): Fix startLockTask throwing

Sevastyan:
#### Summary

`Activity::startLockTask` must be called when `Activity` is in the foreground. Otherwise, the call will fail. This PR ensures that one is able to call the method right on the app start-up - when `Activity` is not drawn (is in the foreground) yet - and get the expected behaviour of app being put into kiosk mode.

#### Testing steps

Test that the app is put into kiosk mode.

#### Follow-up issues

No.

#### Check during review

- Unused code removed.
- Build passing.

### DIFF
--- a/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt
+++ b/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt
@@ -2,6 +2,7 @@ package com.mews.kiosk_mode
 
 import androidx.annotation.NonNull
 import android.app.Activity
+import android.view.ViewGroup
 
 import io.flutter.embedding.engine.plugins.FlutterPlugin
 import io.flutter.embedding.engine.plugins.activity.ActivityAware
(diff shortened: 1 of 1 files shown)


kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt
```diff
@@ -23,7 +24,13 @@ class KioskModePlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
 
     override fun onMethodCall(@NonNull call: MethodCall, @NonNull result: Result) {
         if (call.method == "startKioskMode") {
-            activity?.startLockTask()
+            activity?.let { a ->
```
ookami-kb, line 27: Since it's pretty crucial functionality, let's be more on the safe side:

- method can return bool;
- If activity is null, return false immediately;
- catch `IllegalArgumentException` and return false;
- return true when everything is ok.

ookami-kb (changes requested):

Sevastyan:
@ookami-kb Do we need to adjust our client somehow?

ookami-kb:
It's non-breaking change, but yeah, we should take the result into account.

ookami-kb (approved):

PR #115, opened 2021-07-21
fix(kiosk_mode): Fix startKioskMode signature

Sevastyan:
#### Summary

Changed the signature of the `startKioskMode` method to return `bool` value.

#### Testing steps

No.

#### Follow-up issues

No.

#### Check during review

- Unused code removed.
- Build passing.


kiosk_mode/lib/kiosk_mode.dart
```diff
@@ -7,9 +7,12 @@ const MethodChannel _channel = MethodChannel('com.mews.kiosk_mode/kiosk_mode');
 /// Request to put this activity in a mode where the user is locked to a
 /// restricted set of applications.
 ///
+/// Returns true, if platform satisfied the request successfully,
+/// false - otherwise.
+///
 /// It's only supported on Android currently, where it calls [startLockTask].
 /// It will throw [FlutterMethodNotImplemented] on iOS.
-Future<void> startKioskMode() => _channel.invokeMethod('startKioskMode');
+Future<bool?> startKioskMode() => _channel.invokeMethod<bool>('startKioskMode');
```
ookami-kb, line 15: It should return `bool`, not `bool?`. Let's make a fallback to `false` at least.

ookami-kb (changes requested):

ookami-kb (approved):

PR #124, opened 2021-08-12
feat: Add padding to expansion tile

Gladkowski:
#### Summary

Add padding to expansion_tile.dart

#### Check during review

- Unused code removed.
- Build passing.

### DIFF
--- a/optimus/lib/src/expansion/expansion_tile.dart
+++ b/optimus/lib/src/expansion/expansion_tile.dart
@@ -18,6 +18,7 @@ class OptimusExpansionTile extends StatefulWidget {
     this.leading,
     required this.title,
     this.subtitle,
+    this.contentPadding,
     this.backgroundColor,
     this.onExpansionChanged,
     this.children = const <Widget>[],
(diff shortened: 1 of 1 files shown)


chaeMil (approved):

ookami-kb:
What's the use case? Do we need to have it customizable somewhere?

Gladkowski:
> What's the use case? Do we need to have it customizable somewhere?

@ookami-kb It's going to be used in housekeeping screen redesign. We are defining padding for `ListTile` there => it would be handy if we can change padding for `ExpansionTile` too

ookami-kb:
Do we have it different throughout the app? Isn't this padding strictly defined by design library?

Gladkowski:
> Do we have it different throughout the app? Isn't this padding strictly defined by design library?

good point, tbh, not sure if it is different. The thing is - we don't have `OptimusListTile` in library's documentation, it's more like a specific mobile component. I'll check if design differs throughout the app.

ookami-kb:
@Gladkowski any update on that?

PR #125, opened 2021-08-16
feat!: Update chat avatars support

chaeMil:
#### Summary

Added boolean to toggle the avatar widget. Its used not for hiding but for enabling the avatar support.

This PR is required for the [chat PR](https://github.com/MewsSystems/mews-mobile-commander/pull/765) to fit the design correctly. 

Currently we are supporting just hiding of the avatar widget, but there's always a blank space or the avatar. This adds support for disabling the avatar even with then blank space.

Correct design:
image

Current implementation without the avatars:
image

#### Testing steps

#### Follow-up issues

#### Check during review

- Unused code removed.
- Build passing.

### DIFF
--- a/optimus/lib/src/chat/bubble.dart
+++ b/optimus/lib/src/chat/bubble.dart
@@ -187,12 +187,10 @@ class OptimusChatBubble extends StatelessWidget {
           ? OptimusStackAlignment.start
           : OptimusStackAlignment.end;
 
-  Widget get _avatar => isAvatarEnabled
-      ? SizedBox(
-          width: 40,
-          child: isAvatarVisible ? message.avatar : null,
-        )
-      : Container();
+  Widget get _avatar => SizedBox(
+        width: 40,
+        child: isAvatarVisible ? message.avatar : null,
+      );
 
   BoxDecoration _buildMessageBackground(OptimusThemeData theme) =>
       BoxDecoration(
@@ -237,9 +235,11 @@ class OptimusChatBubble extends StatelessWidget {
           mainAxisAlignment: _bubbleAlignment,
           crossAxisAlignment: OptimusStackAlignment.end,
           children: [
-            if (message.alignment == MessageAlignment.left) _avatar,
+            if (message.alignment == MessageAlignment.left && isAvatarEnabled)
+              _avatar,
             _buildMessageBubble(theme),
-            if (message.alignment == MessageAlignment.right) _avatar,
+            if (message.alignment == MessageAlignment.right && isAvatarEnabled)
+              _avatar,
           ],
         ),
         if (isStatusVisible) _buildStatus(theme),
--- a/optimus/lib/src/chat/bubble.dart
+++ b/optimus/lib/src/chat/bubble.dart
@@ -13,6 +13,7 @@ class OptimusChatBubble extends StatelessWidget {
     Key? key,
     required this.message,
     required this.isAvatarVisible,
+    required this.isAvatarEnabled,
     required this.isStatusVisible,
     required this.isUserNameVisible,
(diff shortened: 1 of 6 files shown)


optimus/lib/src/chat/bubble.dart
```diff
@@ -177,10 +187,12 @@ class OptimusChatBubble extends StatelessWidget {
...
-  Widget get _avatar => SizedBox(
-        width: 40,
-        child: isAvatarVisible ? message.avatar : null,
-      );
+  Widget get _avatar => isAvatarEnabled
+      ? SizedBox(
+          width: 40,
+          child: isAvatarVisible ? message.avatar : null,
+        )
+      : Container();
```
Gladkowski, line 195: I would revert this change to 
```
Widget get _avatar => SizedBox(
        width: 40,
        child: isAvatarVisible ? message.avatar : null,
      );
```

And add the check to `line:240` and `line:242` like:
`if (message.alignment == MessageAlignment.left && isAvatarVisible) _avatar,`

Gladkowski (changes requested):

optimus/lib/src/chat/bubble.dart
chaeMil: Yep looks better :)

Gladkowski (approved):

optimus/lib/src/chat/chat.dart
```diff
@@ -24,6 +25,7 @@ class OptimusChat extends StatelessWidget {
   }
 
   final List<OptimusMessage> _messages = [];
+  final bool isAvatarEnabled;
```
ookami-kb, line 28: Maybe something like `hasAvatars` would be better?

optimus/lib/src/chat/chat.dart
```diff
@@ -41,6 +43,7 @@ class OptimusChat extends StatelessWidget {
               reverse: true,
               itemBuilder: (context, index) => OptimusChatBubble(
                 message: _messages[index],
+                isAvatarEnabled: isAvatarEnabled,
```
ookami-kb, line 46: This looks messy with chat bubble controlling padding based on avatar.

Let's rather make the chat component control avatar rendering and padding, and bubble will render the rest.

ookami-kb (changes requested):

optimus/lib/src/chat/chat.dart
```diff
@@ -39,24 +41,51 @@ class OptimusChat extends StatelessWidget {
...
+                    _avatar(index),
+                ],
               ),
             ),
           ),
           OptimusChatInput(onSendPressed: onSendPressed),
         ],
       );
 
+  Widget _avatar(int index) => Padding(
```
ookami-kb, line 76: `_buildAvatar`

optimus/lib/src/chat/chat.dart
```diff
@@ -39,24 +41,51 @@ class OptimusChat extends StatelessWidget {
...
-                formatDate: formatDate,
-                sending: sending,
-                sent: sent,
-                error: error,
+              itemBuilder: (context, index) => OptimusStack(
+                direction: Axis.horizontal,
+                crossAxisAlignment: OptimusStackAlignment.end,
+                mainAxisAlignment: _bubbleAlignment(index),
+                children: [
+                  if (_messages[index].alignment == MessageAlignment.left &&
```
ookami-kb, line 49: Nitpicking: `hasAvatars && ...` would be slightly more readable – `hasAvatars` sounds more important, if there's no avatar, other checks don't make sense.

ookami-kb (changes requested):

ookami-kb (approved):

PR #128, opened 2021-08-24
feat!(optimus): Update chat functionality and design [ignore-size]

chaeMil:
#### Summary

Added padding to bottom of chat bubble, to prevent messy layouts. 
Also moved the status text outside the message bubble to the chat widget itself. 
Fixed consecutive messages display.
Updated how status is rendered.
Added `currentUserName` attribute for distinguishing between messages.

Needed for [this](https://github.com/MewsSystems/mews-mobile-commander/pull/765).

<details>
  <summary>Screenshots</summary>
  
Before: 
Screenshot 2021-08-26 at 13 03 05

Figma:
Screenshot 2021-08-26 at 13 03 27

Implementation in storybook:
Screenshot 2021-08-26 at 13 02 55

</details>

#### Testing steps

#### Follow-up issues

#### Check during review

- Unused code removed.
- Build passing.


optimus/lib/src/chat/chat.dart
```diff
@@ -74,7 +74,7 @@ class OptimusChat extends StatelessWidget {
       );
 
   Widget _buildAvatar(int index) => Padding(
-        padding: EdgeInsets.only(bottom: _showAvatar(index) ? 18 : 0),
+        padding: EdgeInsets.only(bottom: _showAvatar(index) ? 34 : 0),
```
ookami-kb, line 77: Where does this `34` come from?

chaeMil: 16 from the status padding and then plus its current 18. I know its a magic number :/

ookami-kb: I think the layout should rather be updated. Basically, avatar is bottom-aligned with the message, so it can be something like this:

```
-------- -------------
Avatar   | Message            <- row
-------- -------------
              | Timestamp
```

I guess, currently `18` is coming from the timestamp height, so it will be broken if use has non-default font size.

chaeMil: So the only option to do this is move the status text to chat widget outside of the bubble.

Sevastyan (changes requested):
No PR description - no review.

Sevastyan (approved):

optimus/lib/src/chat/chat.dart
```diff
@@ -32,84 +37,202 @@ class OptimusChat extends StatelessWidget {
   final Widget sent;
   final Widget error;
   final SendCallback onSendPressed;
+  final bool Function(OptimusMessage) isFromCurrentUser;
```
ookami-kb, line 40: Nitpicking: can be changed to `final Predicate<OptimusMessage> isFromCurrentUser` from `dfunc`.

optimus/lib/src/chat/chat.dart
```diff
@@ -1,7 +1,11 @@
 import 'package:flutter/widgets.dart';
 import 'package:optimus/optimus.dart';
-import 'package:optimus/src/chat/input.dart';
 import 'package:optimus/src/chat/message.dart';
+import 'package:optimus/src/chat/optimus_chat_input.dart';
```
ookami-kb, line 4: No padding: 
Screenshot 2021-08-30 at 12 04 40

ookami-kb (changes requested):

ookami-kb (approved):
Currently, the chat interface looks inconsistent to me (in terms of parameters, avatars/username displaying etc). Merging it for now to update in the app, but we'll need to refactor it later.

chaeMil:
> Currently, the chat interface looks inconsistent to me (in terms of parameters, avatars/username displaying etc). Merging it for now to update in the app, but we'll need to refactor it later.

I will create task for this

PR #131, opened 2021-08-30
clean(optimus): [RND-73527] Make OptimusChat API consistent

chaeMil:
#### Summary

Move `avatar` into `OptimusMessageAuthor` which were renamed since it was colliding with our models in `Mobile Commander`

#### Testing steps

#### Follow-up issues

#### Check during review

- Unused code removed.
- Build passing.

### DIFF
--- a/optimus/lib/src/chat/message.dart
+++ b/optimus/lib/src/chat/message.dart
@@ -26,20 +26,20 @@ typedef TryAgainCallback = Future<MessageState> Function(
 @freezed
 class OptimusMessage with _$OptimusMessage {
   const factory OptimusMessage({
-    required MessageAuthor author,
+    required OptimusMessageAuthor author,
     required String message,
     required MessageAlignment alignment,
     required MessageColor color,
     required DateTime time,
     required MessageState state,
-    required Widget? avatar,
   }) = _Message;
 }
 
 @freezed
-class MessageAuthor with _$MessageAuthor {
-  const factory MessageAuthor({
+class OptimusMessageAuthor with _$OptimusMessageAuthor {
+  const factory OptimusMessageAuthor({
     required String id,
     required String username,
-  }) = _MessageAuthor;
+    Widget? avatar,
+  }) = _OptimusMessageAuthor;
 }
--- a/optimus/lib/src/chat/message.freezed.dart
+++ b/optimus/lib/src/chat/message.freezed.dart
@@ -17,13 +17,12 @@ class _$OptimusMessageTearOff {
   const _$OptimusMessageTearOff();
 
   _Message call(
-      {required MessageAuthor author,
+      {required OptimusMessageAuthor author,
       required String message,
       required MessageAlignment alignment,
       required MessageColor color,
       required DateTime time,
-      required MessageState state,
-      required Widget? avatar}) {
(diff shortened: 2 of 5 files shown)


optimus/lib/src/chat/message.dart
```diff
@@ -26,20 +26,20 @@ typedef TryAgainCallback = Future<MessageState> Function(
...
 
 @freezed
-class MessageAuthor with _$MessageAuthor {
-  const factory MessageAuthor({
+class OptimusMessageAuthor with _$OptimusMessageAuthor {
+  const factory OptimusMessageAuthor({
     required String id,
     required String username,
-  }) = _MessageAuthor;
+    Widget? avatar,
```
Sevastyan, line 43: First of all, I don't like that some Widget is included into something that's basically a model class.

Second of all, I'd rather have a narrower API and make it `OptimusAvatar` instead of `Widget`.

All in all, I suggest defining a function on `OptimusChatBubble` like so:

```
OptimusAvatar Function(OptimusMessageAuthor) buildAvatar,
```

Sevastyan (changes requested):

optimus/lib/src/chat/message.freezed.dart
```diff
@@ -17,21 +17,19 @@ class _$OptimusMessageTearOff {
   const _$OptimusMessageTearOff();
 
   _Message call(
-      {required MessageAuthor author,
```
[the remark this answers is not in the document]
chaeMil: Its a library, so they are supposed to be there. Always been there. CC @ookami-kb

optimus/lib/src/chat/message.dart
chaeMil: Ok what about `OptimusMessageAuthor` having nullable `avatarImageUrl` and then build it internally with `OptimusAvatar`?

ookami-kb: This is not some model class from BL, this is a view model, so it's ok to contain a widget. Take a look at [this](https://api.flutter.dev/flutter/widgets/BottomNavigationBarItem-class.html) for example.

chaeMil: So should i revert it?

Sevastyan: A view model with `String id`? Looks messy to me.

ookami-kb: What's messy about it?

chaeMil: Its id of the user. We are using string ids for users in `Mobile Commander` and those are usually UUIDS now, so i dont get whats the problem with that.

ookami-kb: > So should i revert it?

Yes.

Sevastyan: It's messy in terms that it is really a model, albeit a view model. Why creating a Widget in advance, when it might not even be rendered?

ookami-kb: If it's not needed, it won't be rendered. Widget classes are lightweight.

Sevastyan (approved):

ookami-kb (approved):

PR #146, opened 2021-09-16
feat(optimus): Add loader to input field

chaeMil:
#### Summary

Added loaded to input field
Required for this [PR](https://github.com/MewsSystems/mews-mobile-commander/pull/868)

#### Testing steps

Should display loader after suffix when enabled

#### Follow-up issues

None

#### Check during review

- Unused code removed.
- Build passing.


Gladkowski (approved):

optimus/lib/src/input_field.dart
```diff
@@ -100,6 +101,9 @@ class OptimusInputField extends StatefulWidget {
   /// {@macro flutter.widgets.editableText.showCursor}
   final bool? showCursor;
 
+  /// If true, displays [OptimusCircleLoader] after the suffix
+  final bool? showLoader;
```
ookami-kb, line 105: Hm, I'm not sure that we need a separate parameter for that. Isn't it what `suffix` is intended for?

chaeMil: Yeah, can be, but i think that this will simplify using the component by just providing boolean and not placing this in the suffix: `const OptimusCircleLoader(size: OptimusCircleLoaderSize.small, variant: OptimusCircleLoaderVariant.indeterminate() );` when you just need a loader. And when we update design (as we did already) the user dont need to change it everywhere.

ookami-kb: It makes sense. But should it be **near** the suffix or **instead** of it? We can leave the parameter for the convenience, but replace the suffix if `showLoader` is set to `true`. At least, I don't see any examples in the specifications, where we have both loader and suffix.

optimus/lib/src/input_field.dart
```diff
@@ -144,10 +148,16 @@ class _OptimusInputFieldState extends State<OptimusInputField>
                 widget.onChanged?.call('');
               },
             ),
-          if (widget.suffix != null) widget.suffix!,
+          if (widget.showLoader) _loader,
+          if (!widget.showLoader && widget.suffix != null) widget.suffix!,
```
ookami-kb, line 152: What about something like

```dart
if (widget.showLoader) _loader else if (widget.suffix != null) widget.suffix!,
```

ookami-kb (approved):

PR #147, opened 2021-09-16
feat: Update dart metrics

ookami-kb:
#### Summary

Updated analysis options and updated packages to comply with these options.

#### Testing steps

No, code style update.

#### Follow-up issues

No.

#### Check during review

- Unused code removed.
- Build passing.


storybook/lib/stories/search_field.dart
```diff
@@ -54,6 +56,7 @@ class _SearchStoryState extends State<SearchStory> {
             ),
           )
           .toList(),
+      // ignore: prefer-extracting-callbacks
```
Gladkowski, line 59: Do we really need `prefer-extracting-callbacks` in `analysis_options`? I don't think that it adds any value.

Gladkowski (changes requested):

storybook/lib/stories/search_field.dart
ookami-kb: Up to discussion, but I think it makes sense – inline callbacks really make the code less readable.

chaeMil (approved):

PR #151, opened 2021-09-21
fix(optimus): Fix number picker

Gladkowski:
#### Summary

Fix number picker

#### Testing steps

- in storybook, go to `Number Picker` story
- number picker constraints are hardcoded to `[-5; 5]`
- tap on input and type negative value out of range (example: -9)
- errors will appear in console

#### Check during review

- Unused code removed.
- Build passing.

### DIFF
--- a/optimus/lib/src/number_picker.dart
+++ b/optimus/lib/src/number_picker.dart
@@ -67,7 +67,13 @@ class _OptimusNumberPickerState extends State<OptimusNumberPicker> {
   }
 
   void _updateInputValue(int? value) {
-    _controller.value = _controller.value.copyWith(text: _format(value));
+    final formattedValue = _format(value);
+    _controller.value = _controller.value.copyWith(
+      text: formattedValue,
+      selection: TextSelection.fromPosition(
+        TextPosition(offset: formattedValue.length),
+      ),
+    );
   }
 
   void _onMinusTap() {
--- a/optimus/example/pubspec.lock
+++ b/optimus/example/pubspec.lock
@@ -112,7 +112,7 @@ packages:
       name: dart_code_metrics
       url: "https://pub.dartlang.org"
     source: hosted
-    version: "4.3.0"
+    version: "4.3.2"
   dart_style:
     dependency: transitive
     description:
@@ -220,14 +220,14 @@ packages:
       name: mews_pedantic
       url: "https://pub.dartlang.org"
     source: hosted
-    version: "0.4.0-dev.0"
+    version: "0.4.0-dev.1"
   optimus:
     dependency: "direct main"
     description:
       path: ".."
       relative: true
     source: path
-    version: "0.11.2"
+    version: "0.11.5"
   package_config:
     dependency: transitive
     description:
--- a/optimus/pubspec.lock
+++ b/optimus/pubspec.lock
@@ -353,7 +353,7 @@ packages:
       name: mews_pedantic
       url: "https://pub.dartlang.org"
     source: hosted
-    version: "0.4.0-dev.0"
+    version: "0.4.0-dev.1"
   mime:
     dependency: transitive
     description:
--- a/storybook/pubspec.lock
+++ b/storybook/pubspec.lock
@@ -112,7 +112,7 @@ packages:
       name: dart_code_metrics
       url: "https://pub.dartlang.org"
     source: hosted
-    version: "4.3.0"
+    version: "4.3.1"
   dart_style:
     dependency: transitive
     description:
@@ -234,7 +234,7 @@ packages:
       name: mews_pedantic
       url: "https://pub.dartlang.org"
     source: hosted
-    version: "0.4.0-dev.0"
+    version: "0.4.0-dev.1"
   nested:
     dependency: transitive
     description:
@@ -248,7 +248,7 @@ packages:
       path: "../optimus"
       relative: true
     source: path
-    version: "0.11.2"
+    version: "0.11.5"
   package_config:
     dependency: transitive
     description:


chaeMil (approved):

optimus/lib/src/number_picker.dart
```diff
@@ -67,7 +67,13 @@ class _OptimusNumberPickerState extends State<OptimusNumberPicker> {
   }
 
   void _updateInputValue(int? value) {
-    _controller.value = _controller.value.copyWith(text: _format(value));
+    final formattedValue = _format(value);
+    _controller.value = _controller.value.copyWith(
+      text: formattedValue,
+      selection: TextSelection.fromPosition(
```
ookami-kb, line 73: With that change, you cannot type something into the middle. E.g. you have `200` and you want to turn it into `23400`. After pressing `3` your cursor will be at the end.

ookami-kb (changes requested):

chaeMil (approved):
Yep works better now :)

ookami-kb (changes requested):
1. Enter `100`.
2. Select last `0`.
3. Press `-` button.
4. Observe error.

You need to check that the whole selection makes sense. Otherwise, just put the cursor at the end.

chaeMil (approved):
For me works as expected, no errors, tried all possible variants.

optimus/lib/src/number_picker.dart
```diff
@@ -56,18 +56,29 @@ class _OptimusNumberPickerState extends State<OptimusNumberPicker> {
...
-  void _updateInputValue(int? value) {
-    _controller.value = _controller.value.copyWith(text: _format(value));
+  void _updateInputValue({int? value, bool isInRange = true}) {
+    final formattedValue = _format(value);
+    final length = formattedValue.length;
+    final selection = _controller.value.selection;
+
+    _controller.value = _controller.value.copyWith(
+      text: formattedValue,
+      selection: isInRange &&
```
ookami-kb, line 76: Why do you care about `isInRange` at all?

Gladkowski: Let's say max value is 999.
Type `123` and put cursor before `1`.
Press any number on keyboard, the selection will be at 1. Press any number again the selection will move to 2. And so on. That's why we need to know if value is not in range to reset the selection to values length.

ookami-kb (changes requested):
1. Follow the structure we already have: `*FormField` component should be in a different folder.
2. If I erase the value, it shows an error. Previously, it gave me default value. Is it discussed with designers?

Gladkowski:
> 1. Follow the structure we already have: `*FormField` component should be in a different folder.
> 2. If I erase the value, it shows an error. Previously, it gave me default value. Is it discussed with designers?

Confirmed with designers: the widget should have default value

ookami-kb:
@Gladkowski conflicts

optimus/lib/src/number_picker/number_picker.dart
```diff
@@ -0,0 +1,174 @@
...
+    required this.onChanged,
+    this.initialValue = 0,
+    this.defaultValue = 0,
+    this.min = 0,
+    this.max = 100,
+    this.focusNode,
+    this.enabled = true,
+    this.error,
+    this.readOnly = false,
+    this.showCursor,
```
ookami-kb, line 66: These parameters are not used.

optimus/lib/src/number_picker/number_picker.dart
```diff
@@ -0,0 +1,174 @@
...
+          validator: (value) =>
+              value != null && value >= min && value <= max ? null : error,
+          enabled: enabled,
+          autovalidateMode: autovalidateMode,
+          builder: (FormFieldState<int> field) => _OptimusNumberPicker(
+            initialValue: initialValue,
+            defaultValue: defaultValue,
+            min: min,
+            max: max,
+            // ignore: prefer-extracting-callbacks
```
ookami-kb, line 40: Why ignore?

optimus/lib/src/number_picker/number_picker.dart
```diff
@@ -0,0 +1,174 @@
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+import 'package:optimus/optimus.dart';
+import 'package:optimus/src/number_picker/button.dart';
+
+class OptimusNumberPickerFormField extends FormField<int> {
+  OptimusNumberPickerFormField({
+    Key? key,
+    int initialValue = 0,
```
ookami-kb, line 9: Since we allow empty field, initial value can be `null`.

optimus/lib/src/number_picker/number_picker.dart
```diff
@@ -0,0 +1,174 @@
...
+
+  /// {@macro flutter.widgets.editableText.showCursor}
+  final bool? showCursor;
+
+  @override
+  _OptimusNumberPickerState createState() => _OptimusNumberPickerState();
+}
+
+class _OptimusNumberPickerState extends State<_OptimusNumberPicker> {
+  TextEditingController? _controller;
```
ookami-kb, line 89: It can be `late final`.

optimus/lib/src/number_picker/number_picker.dart
```diff
@@ -0,0 +1,174 @@
...
+  FocusNode? _focusNode;
+  late int _value;
```
ookami-kb, line 91: Initialize it here.

ookami-kb (changes requested):

ookami-kb:
@Gladkowski I've pushed some updates

Gladkowski:
> @Gladkowski I've pushed some updates

checked

PR #152, opened 2021-09-30
feat(optimus): Make badge text optional

chaeMil:
#### Summary

Make badge text optional and change rendering of badge based on that.

#### Testing steps

#### Follow-up issues

Needed for [this](https://github.com/MewsSystems/mews-mobile-commander/pull/888)

#### Check during review

- Unused code removed.
- Build passing.

### DIFF
--- a/optimus/lib/src/badge.dart
+++ b/optimus/lib/src/badge.dart
@@ -11,7 +11,7 @@ class OptimusBadge extends StatelessWidget {
 
   final String? text;
 
-  bool get hasText => text != null && text!.isNotEmpty;
+  bool get _hasText => text != null && text!.isNotEmpty;
 
   @override
   Widget build(BuildContext context) {
@@ -27,10 +27,10 @@ class OptimusBadge extends StatelessWidget {
 
     return Container(
       constraints: BoxConstraints(
-        minWidth: hasText ? 20 : 10,
-        maxWidth: hasText ? double.infinity : 10,
+        minWidth: _hasText ? 20 : 10,
+        maxWidth: _hasText ? double.infinity : 10,
       ),
-      height: hasText ? 20 : 10,
+      height: _hasText ? 20 : 10,
       decoration: decoration,
       padding: const EdgeInsets.symmetric(
         horizontal: spacing50,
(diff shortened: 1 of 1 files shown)


chaeMil:
@Gladkowski ping

optimus/lib/src/badge.dart
```diff
@@ -6,10 +6,12 @@ import 'package:optimus/src/theme/theme.dart';
...
   const OptimusBadge({
     Key? key,
-    required this.text,
+    this.text,
   }) : super(key: key);
 
-  final String text;
+  final String? text;
+
+  bool get hasText => text != null && text!.isNotEmpty;
```
Gladkowski, line 14: I would make it private `_hasText`

Gladkowski (changes requested):

Gladkowski (approved):

optimus/lib/src/badge.dart
```diff
@@ -6,10 +6,12 @@ import 'package:optimus/src/theme/theme.dart';
...
   const OptimusBadge({
     Key? key,
-    required this.text,
+    this.text,
   }) : super(key: key);
 
-  final String text;
+  final String? text;
+
+  bool get _hasText => text != null && text!.isNotEmpty;
```
ookami-kb, line 14: Since we don't distinguish between `null` and empty string, make the `text` param non-nullable with empty string by default.

ookami-kb (changes requested):

optimus/lib/src/badge.dart
```diff
@@ -6,11 +6,13 @@ import 'package:optimus/src/theme/theme.dart';
 class OptimusBadge extends StatelessWidget {
   const OptimusBadge({
     Key? key,
-    required this.text,
+    this.text = '',
   }) : super(key: key);
 
   final String text;
 
+  bool get _hasText => text.isNotEmpty;
```
ookami-kb, line 14: Nitpicking: let's make it a local variable inside the `build` method, it's not used anywhere outside.

ookami-kb (approved):

PR #153, opened 2021-10-01
feat: Update pedantic

ookami-kb:
#### Summary

Temporarily disabled `prefer-extracting-callbacks` rule, since it doesn't work properly in the current version of `dart_code_metrics`. It will be fixed in their next release, after that, we can turn it back on.

#### Testing steps

No.

#### Follow-up issues

No.

#### Check during review

- Unused code removed.
- Build passing.

### DIFF
--- a/kiosk_mode/example/pubspec.yaml
+++ b/kiosk_mode/example/pubspec.yaml
@@ -16,6 +16,6 @@ dependencies:
 dev_dependencies:
   flutter_test:
     sdk: flutter
-  mews_pedantic: ^0.4.0-dev.1
+  mews_pedantic: ^0.4.0-dev.2
 flutter:
   uses-material-design: true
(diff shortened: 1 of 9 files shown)


chaeMil (approved):

melos.yaml
```diff
@@ -22,7 +22,7 @@ scripts:
   build_runner: melos exec --scope="optimus" --fail-fast -- "exec flutter packages pub run build_runner build"
 
   test:
-    run: melos exec -- "flutter test --no-pub"
+    run: melos exec -c 1 -- "flutter test --no-pub"
```
Sevastyan, line 25: What does this change do?

ookami-kb: Runs the scripts consequently. It's easier to investigate the logs when they are one by one, and the speed difference is not that big.

PR #170, opened 2021-11-16
feat(optimus)!: Add controller to `OptimusNumberPicker`

chaeMil:
#### Summary

Add controller to `OptimusNumberPicker`. 
Needed for [RND-84283](https://mews.myjetbrains.com/youtrack/issue/RND-84283);

#### Testing steps

The value can be changed from outside the widget (by the controller)

#### Follow-up issues

None

#### Check during review

- Unused code removed.
- Build passing.


Sevastyan:
@chaeMil Please, don't forget to fill in testing steps and follow-up issues sections in PR description.

chaeMil:
> @chaeMil Please, don't forget to fill in testing steps and follow-up issues sections in PR description.

Done

optimus/lib/src/number_picker/number_picker.dart
```diff
@@ -76,25 +77,34 @@ class _OptimusNumberPicker extends StatefulWidget {
...
 }
 
-class _OptimusNumberPickerState extends State<_OptimusNumberPicker> {
-  late final TextEditingController _controller = TextEditingController(
+class _OptimusNumberPickerState extends State<OptimusNumberPicker> {
+  late final _textEditingController = TextEditingController(
     text: widget.initialValue?.toString() ?? '',
   );
   late int _value = widget.initialValue ?? widget.defaultValue;
+  late final NumberPickerController? _numberPickerController =
```
Sevastyan, line 91: Don't make a property `late` and nullable at the same time. It should be either `late`, or nullable and variable. We prefer the later one.

optimus/lib/src/number_picker/number_picker.dart
```diff
@@ -165,3 +180,9 @@ class _OptimusNumberPickerState extends State<_OptimusNumberPicker> {
 }
 
 final _integersOrEmptyString = RegExp(r'^$|^[-]?\d+|^[-]');
+
+class NumberPickerController {
+  late final Function(int value) onSetValue;
```
Sevastyan, line 185: Naming. Better have it as `onValueSet`.

Sevastyan (changes requested):
Describe the intended solution for the issue RND-84283 first, please. Without it, I cannot really judge these changes.

How making `OptimusNumberPicker` public will solve the problem? It is not used anywhere (and should not be used). Adding a constructor parameter to `OptimusNumberPickerFormField` should be done instead to achieve what I think the intention is.

optimus/lib/src/number_picker/number_picker.dart
```diff
@@ -165,3 +180,9 @@ class _OptimusNumberPickerState extends State<_OptimusNumberPicker> {
 }
 
 final _integersOrEmptyString = RegExp(r'^$|^[-]?\d+|^[-]');
+
+class NumberPickerController {
```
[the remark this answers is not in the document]
chaeMil: To set the value from outside the widget.

Sevastyan (approved):

Sevastyan:
@chaeMil "Controller" should do more than just setting a value, i.e. the name implies it has more methods.

ookami-kb (commented):
Why not just pass optional controller instead of doing this "callback of a callback"?

chaeMil:
> Why not just pass optional controller instead of doing this "callback of a callback"?

Honestly im ok with both variants. @Sevastyan

Sevastyan:
@ookami-kb do you mean `TextEditingController`? Then, I guess in order to have a lean API. Widget should only expose the ability of setting the value, from what I got from the issue description.

ookami-kb:
This is a common practice (and in some cases a necessity, actually) to pass the controller to the form field. Take a look at [TextFormField](https://api.flutter.dev/flutter/material/TextFormField-class.html), for example.

Sevastyan:
Let's refactor it then, if it's a common practice.

chaeMil:
@ookami-kb plese check

ookami-kb:
@chaeMil I was talking about passing `TextEditingController` instead of a custom controller.

chaeMil:
@ookami-kb but then there will be problem with the `_value` not being set` without some nasty widget internal work.

ookami-kb:
@chaeMil what do you mean?

chaeMil:
Take a look at lines `115` and `125`. After the press the internal `_value` is changed. But when we change the text with `_textEditingController` the `_value` will not update thus the widget will not work correctly.

chaeMil:
@ookami-kb  ready

optimus/lib/src/number_picker/number_picker.dart
```diff
@@ -94,7 +101,7 @@ class _OptimusNumberPickerState extends State<_OptimusNumberPicker> {
 
   @override
   void dispose() {
-    _controller.dispose();
+    _textEditingController.dispose();
```
ookami-kb, line 104: You cannot just dispose a controller that potentially comes from the widget. Only the entity that creates something is in charge of disposing it. Take a look at focus node.

Also, how does this widget react to controller value update?

ookami-kb (changes requested):

optimus/lib/src/number_picker/number_picker.dart
chaeMil: OK but im still unsure how to deal with that :/

ookami-kb: In the same way as focus node is treated. If it is external controller, do nothing, if internal – dispose it.

Sevastyan (commented):
Shouldn't a controllers listener trigger `_updateValue` ?

```
diff --git a/optimus/lib/src/number_picker/number_picker.dart b/optimus/lib/src/number_picker/number_picker.dart
index 97140fd..abcd9d8 100644
--- a/optimus/lib/src/number_picker/number_picker.dart
+++ b/optimus/lib/src/number_picker/number_picker.dart
@@ -125,7 +125,6 @@ class _OptimusNumberPickerState extends State<_OptimusNumberPicker> {
         : _value > widget.max
             ? widget.max
             : _value - 1;
-    _updateValue(value);
     _updateController(value);
   }
 
@@ -135,7 +134,6 @@ class _OptimusNumberPickerState extends State<_OptimusNumberPicker> {
         : _value > widget.max - 1
             ? widget.max
             : _value + 1;
-    _updateValue(value);
     _updateController(value);
   }
 
```

optimus/lib/src/number_picker/number_picker.dart
```diff
@@ -76,26 +79,43 @@ class _OptimusNumberPicker extends StatefulWidget {
...
-    text: widget.initialValue?.toString() ?? '',
-  );
   late int _value = widget.initialValue ?? widget.defaultValue;
 
+  TextEditingController? _controller;
+
+  TextEditingController get _effectiveController =>
+      widget.controller ??
+      (_controller ??= TextEditingController(
+        text: widget.initialValue?.toString() ?? '',
```
ookami-kb, line 96: Initial value should be passed to controller even if it's external one.

optimus/lib/src/number_picker/number_picker.dart
```diff
@@ -76,26 +79,43 @@ class _OptimusNumberPicker extends StatefulWidget {
...
+
   FocusNode? _focusNode;
 
   FocusNode get _effectiveFocusNode =>
       widget.focusNode ?? (_focusNode ??= FocusNode());
 
+  @override
+  void initState() {
+    super.initState();
+    _effectiveController.addListener(() {
```
ookami-kb, line 107: Listener should be removed from `_effectiveController` on dispose.

optimus/lib/src/number_picker/number_picker.dart
```diff
@@ -76,26 +79,43 @@ class _OptimusNumberPicker extends StatefulWidget {
...
+      final int? value = int.tryParse(_effectiveController.text);
+      if (value != null) {
```
ookami-kb, line 109: I'm not thinking, let's probably remove `defaultValue` and make all the callbacks and `_value` nullable. As it turned out, this "default value" concept is pretty confusing – we should explicitly describe the situation "there's no correct value entered". So it will be up to a caller to fallback to default value if needed.

ookami-kb (changes requested):

optimus/lib/src/number_picker/number_picker.dart
chaeMil: Can we maybe take a look at this together? When trying to accomplish this it became a huge mess so either im not getting it right or it's just a huge mess :D

chaeMil:
ping @Sevastyan @ookami-kb

optimus/lib/src/number_picker/number_picker.dart
```diff
@@ -141,25 +169,20 @@ class _OptimusNumberPickerState extends State<_OptimusNumberPicker> {
           error: widget.error,
           isEnabled: widget.enabled,
           keyboardType: TextInputType.number,
-          controller: _controller,
+          controller: _effectiveController,
           prefix: NumberPickerButton(
             iconData: OptimusIcons.minus_simple,
-            onPressed: _value > widget.min ? _onMinusTap : null,
+            onPressed: (_value ?? widget.max) > widget.min ? _onMinusTap : null,
```
Sevastyan, line 175: `?? widget.max` is logic duplication. It should be:

```dart
onPressed: _value == null || _value! > widget.min ? _onMinusTap : null,
```

to achieve the same behaviour.

optimus/lib/src/number_picker/number_picker.dart
```diff
@@ -141,25 +169,20 @@ class _OptimusNumberPickerState extends State<_OptimusNumberPicker> {
...
           ),
           suffix: NumberPickerButton(
             iconData: OptimusIcons.plus_simple,
-            onPressed: _value < widget.max ? _onPlusTap : null,
+            onPressed: (_value ?? widget.min) < widget.max ? _onPlusTap : null,
```
Sevastyan, line 179:
```dart
onPressed: _value == null || _value! < widget.max ? _onPlusTap : null,
```

Sevastyan (changes requested):

optimus/lib/src/number_picker/number_picker.dart
```diff
@@ -60,73 +60,101 @@ class _OptimusNumberPicker extends StatefulWidget {
...
-    _updateValue(value);
+    late int value;
+    if (_value != null) {
+      value = _value! < widget.min + 1
+          ? widget.min
+          : _value! > widget.max
+              ? widget.max
+              : _value! - 1;
+    } else {
+      value = widget.max;
```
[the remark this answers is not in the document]
chaeMil: I added the opposite ends of the range in plus/minus buttons. It felt natural to me that if theres no value you press minus it will select the highest possible and then you can decrease it same as on the plus button side but in reverse.

optimus/lib/src/number_picker/number_picker.dart
```diff
@@ -7,49 +7,49 @@ class OptimusNumberPickerFormField extends FormField<int> {
...
-          defaultValue >= min && defaultValue <= max,
-          'defaultValue value should be in [min, max] range',
-        ),
         super(
           key: key,
-          initialValue: initialValue ?? defaultValue,
+          initialValue: initialValue,
           onSaved: onSaved,
-          validator: (value) => value != null && value >= min && value <= max
+          validator: (value) => value == null || (value >= min && value <= max)
```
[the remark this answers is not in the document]
chaeMil: Yeah looking at it at retrospectively this doesnt make much sense :D

optimus/lib/src/number_picker/number_picker.dart
```diff
@@ -60,73 +60,101 @@ class _OptimusNumberPicker extends StatefulWidget {
...
   @override
   _OptimusNumberPickerState createState() => _OptimusNumberPickerState();
 }
 
 class _OptimusNumberPickerState extends State<_OptimusNumberPicker> {
-  late final TextEditingController _controller = TextEditingController(
-    text: widget.initialValue?.toString() ?? '',
-  );
-  late int _value = widget.initialValue ?? widget.defaultValue;
+  late int? _value = widget.initialValue;
```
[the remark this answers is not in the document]
chaeMil: We can add assert that setting `initialValue` and `controller` is not allowed.

optimus/lib/src/number_picker/number_picker.dart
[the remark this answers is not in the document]
Sevastyan: I got it. It isn't what I would expect as a user though.

optimus/lib/src/number_picker/number_picker.dart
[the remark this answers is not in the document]
Sevastyan: We can, but we can also delete `initialValue`. Not sure what is considered more appropriate in Flutter world. @ookami-kb

optimus/lib/src/number_picker/number_picker.dart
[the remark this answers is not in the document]
Sevastyan: Since this is a matter of preference, what about pushing this decision to the client?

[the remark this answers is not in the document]
chaeMil: Or we can use `initialValue` if we are not going to get rid of it

optimus/lib/src/number_picker/number_picker.dart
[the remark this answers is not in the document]
ookami-kb: Yes, what's @chaeMil suggested is reasonable, this is how it's done in `TextFormField`:

```dart
  /// When a [controller] is specified, [initialValue] must be null (the
  /// default). If [controller] is null, then a [TextEditingController]
  /// will be constructed automatically and its `text` will be initialized
  /// to [initialValue] or the empty string.
```

optimus/lib/src/number_picker/number_picker.dart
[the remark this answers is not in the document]
ookami-kb: I've checked how it's done in HTML `<input type="number">` and if the value is `null`, both `minus` and `plus` set the `min` value, so we can do the same.

optimus/lib/src/number_picker/number_picker.dart
```diff
@@ -60,73 +65,111 @@ class _OptimusNumberPicker extends StatefulWidget {
...
+    _initValue();
+    _updateController(_value);
+    _effectiveController.addListener(_controllerListener);
+  }
+
   @override
   void dispose() {
-    _controller.dispose();
     _focusNode?.dispose();
+    _controller?.removeListener(_controllerListener);
```
ookami-kb, line 129: Listener should be removed from `_effectiveController`.

optimus/lib/src/number_picker/number_picker.dart
```diff
@@ -60,73 +65,110 @@ class _OptimusNumberPicker extends StatefulWidget {
...
   void _onMinusTap() {
-    final value = _value < widget.min + 1
-        ? widget.min
-        : _value > widget.max
-            ? widget.max
-            : _value - 1;
-    _updateValue(value);
-    _updateController(value);
+    final value = _value;
+    late final int newValue;
```
ookami-kb, line 135: You don't need to use `late` here.

optimus/lib/src/number_picker/number_picker.dart
```diff
@@ -60,73 +65,110 @@ class _OptimusNumberPicker extends StatefulWidget {
...
   void _onPlusTap() {
-    final value = _value < widget.min
-        ? widget.min
-        : _value > widget.max - 1
-            ? widget.max
-            : _value + 1;
-    _updateValue(value);
-    _updateController(value);
+    final value = _value;
+    late final int newValue;
```
ookami-kb, line 150: You don't need to use `late` here.

Sevastyan (approved):
LGTM

optimus/lib/src/number_picker/number_picker.dart
```diff
@@ -60,73 +65,109 @@ class _OptimusNumberPicker extends StatefulWidget {
...
+      _value = widget.initialValue;
+    }
+  }
+
+  @override
+  void initState() {
+    super.initState();
+
+    _initValue();
+    _updateController(_value);
```
ookami-kb, line 121: Since we cannot have both `initialValue` and `controller`, do we need this here?

optimus/lib/src/number_picker/number_picker.dart
```diff
@@ -4,52 +4,57 @@ import 'package:optimus/optimus.dart';
...
               : validationError,
           enabled: enabled,
           autovalidateMode: autovalidateMode,
           builder: (FormFieldState<int> field) {
-            void _onChanged(int value) {
+            void _onChanged(int? value) {
               field.didChange(value);
-              if (onChanged != null && value >= min && value <= max) {
+              if (onChanged != null &&
+                  (value == null || value >= min && value <= max)) {
```
ookami-kb, line 44: I'm not sure about this.

Since we report `null` values (when empty or NaN), shouldn't we also report out-of-range values either as `null`? WDYT?

ookami-kb (changes requested):

optimus/lib/src/number_picker/number_picker.dart
chaeMil: Nope, but anyway i discovered a bug now. When then `controller` is set and have some value the `validation error` is shown.

chaeMil: @ookami-kb Can be fixed by adding 
```
WidgetsBinding.instance!.addPostFrameCallback((_) {
      widget.onChanged(_value);
 });
```
into `initState` but i dont like it

chaeMil: This will fix the validation error popping but for some reason the `_value` is still null.

chaeMil:
@ookami-kb Dont ask me, it works like this...

optimus/lib/src/number_picker/number_picker.dart
Sevastyan: As a user, I'd prefer if I'm not disturbed with out-of-range values on my `onChanged` callback. I might have logic connected to values that are `null` (field empty, or NaN), so reporting some out-of-range *number* as `null` is not what I expect. Out-of-range should not happen, so it should not be reported.

But if we want to be precise, I would introduce a union and report it to `onChanged`:

```
number
out-of-range number
null
NaN
```

And that is an over-engineering.

chaeMil: This actually seems like a nice solution. WDYT @ookami-kb?

chaeMil: ping @ookami-kb so how about adding this union?

ookami-kb: AS for me, this union doesn't make much sense.

We need to track the fact that out-of-range value was entered (e.g. in Commander the price shouldn't stay the same when changing the amount from 50 to 500, otherwise it would be confusing), but we don't really need the exact reason why the value is invalid (and it will make the usage more complex since you would have to unwrap the value).

If the exact invalid value is needed, you can just provide the controller and get the raw entered value.

chaeMil: So lets just call `onChange` on every change even if its out of range and the rest leave up to user to use provided controller, right?

chaeMil:
@ookami-kb yeah nice idea :)

ookami-kb (approved):

PR #187, opened 2022-03-03
fix(optimus): RenderBox.size: Null check operator used on a null value

chaeMil:
#### Summary

Did the same check as in other places regarding `RenderBox`, but here if its `null` returns the previous `_savedRect` 🤷‍♂️

#### Testing steps

None

#### Follow-up issues

None

#### Check during review

- Unused code removed.
- Build passing.

### DIFF
--- a/optimus/lib/src/common/dropdown.dart
+++ b/optimus/lib/src/common/dropdown.dart
@@ -94,11 +94,12 @@ class _OptimusDropdownState<T> extends State<OptimusDropdown<T>>
   }
 
   Rect _calculateRect() {
-    final RenderBox renderBox =
-        widget.anchorKey.currentContext?.findRenderObject() as RenderBox;
-    final size = renderBox.size;
+    final renderObject = widget.anchorKey.currentContext?.findRenderObject();
+    if (renderObject is! RenderBox || !renderObject.hasSize) return _savedRect;
 
-    return renderBox.localToGlobal(Offset.zero) & size;
+    final size = renderObject.size;
+
+    return renderObject.localToGlobal(Offset.zero) & size;
   }
 
   Widget _buildListView(bool isReversed) => Material(
--- a/optimus/lib/src/common/dropdown.dart
+++ b/optimus/lib/src/common/dropdown.dart
@@ -26,12 +26,13 @@ class OptimusDropdown<T> extends StatefulWidget {
 
 class _OptimusDropdownState<T> extends State<OptimusDropdown<T>>
     with ThemeGetter {
-  late Rect _savedRect = _calculateRect();
+  // shouldn't be null since its called right after widget creation
+  late Rect _savedRect = _calculateRect()!;
 
   void _updateRect(dynamic _) {
     if (!mounted) return;
     final newRect = _calculateRect();
-    if (newRect != _savedRect) {
+    if (newRect != null && newRect != _savedRect) {
       setState(() {
         _savedRect = newRect;
       });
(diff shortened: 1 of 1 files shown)


optimus/lib/src/common/dropdown.dart
```diff
@@ -94,11 +94,12 @@ class _OptimusDropdownState<T> extends State<OptimusDropdown<T>>
   }
 
   Rect _calculateRect() {
-    final RenderBox renderBox =
-        widget.anchorKey.currentContext?.findRenderObject() as RenderBox;
-    final size = renderBox.size;
+    final renderObject = widget.anchorKey.currentContext?.findRenderObject();
+    if (renderObject is! RenderBox || !renderObject.hasSize) return _savedRect;
```
ookami-kb, line 98: `_savedRect` uses `_calculateRect` for initial value. It can lead to the infinite recursive call.

ookami-kb (changes requested):

ookami-kb:
@chaeMil any progress?

optimus/lib/src/common/dropdown.dart
```diff
@@ -26,12 +26,13 @@ class OptimusDropdown<T> extends StatefulWidget {
 
 class _OptimusDropdownState<T> extends State<OptimusDropdown<T>>
     with ThemeGetter {
-  late Rect _savedRect = _calculateRect();
+  // shouldn't be null since its called right after widget creation
```
ookami-kb, line 29: Why in that case it should become `null` later?

chaeMil: @ookami-kb ping

ookami-kb: I don't like these implicit assumptions about "shouldn't be null". Can we initialize it to let's say empty `Rect`? Will it have some visual glitches?

chaeMil: I think that the pop up part will go above (in term of z-axis) the button it self and cover it. I will try ;)

chaeMil: image
Basically it will cover the button itself. Dont mind the offset in storybook its like this always. We can probably do fallback like this.

chaeMil:
ping @ookami-kb

optimus/lib/src/common/dropdown.dart
ookami-kb: This is how it looks for me:

image

> Dont mind the offset in storybook its like this always.

It's really weird and should be fixed in that case.

chaeMil: Ok i will look into it, but the solution with empty `Rect` could work right?

chaeMil: Yeaht it looks the same for me without the empty `Rect`

ookami-kb: Yep, I think empty Rect should work. Let's just create a separate issue for incorrect rendering in the storybook.

optimus/lib/src/common/dropdown.dart
```diff
@@ -26,6 +26,7 @@ class OptimusDropdown<T> extends StatefulWidget {
 
 class _OptimusDropdownState<T> extends State<OptimusDropdown<T>>
     with ThemeGetter {
+  // shouldn't be null since its called right after widget creation
```
ookami-kb, line 29: This comment is now obsolete.

ookami-kb (changes requested):

ookami-kb (approved):

PR #204, opened 2022-05-17
feat(optimus): [RND-103047] Added OptimusNotification

witwash:
#### Summary

RND-103047

Added OptimusNotification.

<details>
<summary>Screenshots</summary>
Screenshot 2022-05-17 at 10 36 38
Screenshot 2022-05-17 at 10 36 47
Screenshot 2022-05-17 at 10 36 52
Screenshot 2022-05-17 at 10 36 56
Screenshot 2022-05-17 at 10 37 15

</details>

#### Testing steps

#### Follow-up issues

The next step will be add stacking functionality to notification component (RND-104514).

#### Check during review

- Unused code removed.
- Build passing.


optimus/lib/src/notification.dart
```diff
@@ -0,0 +1,229 @@
...
+import 'package:optimus/src/elevation.dart';
+import 'package:optimus/src/typography/presets.dart';
+
+const double _kMaxWidth = 360;
+const double _kCloseIconSize = 16;
+const double _kIconSize = 20;
+const double _kIconHorizontalPadding = 10;
+const int _kMaxLinesTitle = 1;
+const int _kMaxLinesDescription = 5;
+const int _kMaxLinesLink = 1;
```
Sevastyan, line 14: I don't think we should use `_k` prefix.
@ookami-kb what's the convention in this project on this?

optimus/lib/src/notification.dart
```diff
@@ -0,0 +1,229 @@
...
+      case OptimusNotificationVariant.warning:
+        return theme.colors.neutral1000;
+    }
+  }
+}
+
+/// Notification are used for showing brief and concise message that
+/// communicates immediate feedback with optional action included. Notifications
+/// are noticeable but not intrusive to the use and can be temporary.
+class OptimusNotification extends StatelessWidget {
```
Sevastyan, line 75: Why the most important piece starts on line `75`? Can we move enums and extensions to the end of the file?

optimus/lib/src/notification.dart
```diff
@@ -0,0 +1,229 @@
...
+      case OptimusNotificationVariant.success:
+      case OptimusNotificationVariant.danger:
+        return theme.colors.neutral0;
+      case OptimusNotificationVariant.warning:
+        return theme.colors.neutral1000;
+    }
+  }
+}
+
+/// Notification are used for showing brief and concise message that
```
Sevastyan, line 72: Typo: "Notification **is**".

optimus/lib/src/notification.dart
```diff
@@ -0,0 +1,229 @@
...
+/// communicates immediate feedback with optional action included. Notifications
+/// are noticeable but not intrusive to the use and can be temporary.
+class OptimusNotification extends StatelessWidget {
+  const OptimusNotification({
+    Key? key,
+    required this.title,
+    this.icon,
+    this.link,
+    this.onLinkPressed,
+    this.description,
```
Sevastyan, line 82: Nitpick: I'd rather name it `body` so it goes nicely with `title`.

```
title
body
```

optimus/lib/src/notification.dart
```diff
@@ -0,0 +1,229 @@
...
+    this.isDismissible = false,
+    this.onDismiss,
```
Sevastyan, line 84: Nitpick.
`onLinkPressed` is in the past tense, but `onDismiss` is not. Let's make it more consistent with `onDismissed`.

Sevastyan (changes requested):
What about extracting most important bits of a build Widget method into functions (or Widgets) to aid readability?

g964

optimus/lib/src/notification.dart
witwash: I guess you are right. I saw it in a few places (`expansion_tile.dart` and `optimus_icons.dart`), but most constants are named without `_k`: `spacing.dart` `border_radius.dart` etc.

optimus/lib/src/notification.dart
witwash: I left the notification variant enum at the top, but extensions are at the bottom now.

optimus/lib/src/notification.dart
```diff
@@ -0,0 +1,349 @@
+import 'dart:math';
+import 'package:flutter/material.dart';
+import 'package:optimus/optimus.dart';
+import 'package:optimus/src/border_radius.dart';
+import 'package:optimus/src/elevation.dart';
+import 'package:optimus/src/typography/presets.dart';
+
+/// Describes a certain type of notification with its semantical meaning.
```
ookami-kb, line 8: Add blank line after the first sentence.

optimus/lib/src/notification.dart
```diff
@@ -0,0 +1,349 @@
...
+///  - [OptimusNotificationVariant.danger] - Used for notifying about the
+/// dangerous matter. Could be error, destructive action or negative feedback.
+enum OptimusNotificationVariant {
+  info,
+  success,
+  warning,
+  danger,
+}
+
+///Notification is used for showing a brief and concise message that
```
ookami-kb, line 25: Add space between `///` and text.

optimus/lib/src/notification.dart
```diff
@@ -0,0 +1,349 @@
...
+    this.variant = OptimusNotificationVariant.info,
+  }) : super(key: key);
+
+  final String title;
+  final String? body;
+  final IconData? icon;
+  final String? link;
+  final VoidCallback? onLinkPressed;
+  final bool isDismissible;
+  final VoidCallback? onDismissed;
```
ookami-kb, line 47: Let's use the same approach as in default button – if it's dismissible, `onDismissed` shouldn't be `null`. That means that `isDismissible` parameter is not needed.

optimus/lib/src/notification.dart
```diff
@@ -0,0 +1,349 @@
...
+            size: _closeIconSize,
+          ),
+        ),
+      ),
+    );
+  }
+}
+
+extension on Breakpoint {
+  double getPadding() {
```
ookami-kb, line 291: Use getter instead.

optimus/lib/src/notification.dart
```diff
@@ -0,0 +1,349 @@
...
+      case OptimusNotificationVariant.success:
+        return theme.colors.success500;
+      case OptimusNotificationVariant.warning:
+        return theme.colors.warning500;
+      case OptimusNotificationVariant.danger:
+        return theme.colors.danger500;
+    }
+  }
+
+  IconData getBannerIcon() {
```
ookami-kb, line 318: Use getter.

ookami-kb (changes requested):

optimus/lib/src/notification.dart
witwash: Do you mean to replace the entire extension on `OptimusNotificatioVariant` and move them to getters or just that one?

ookami-kb: I mean just changing this one to `IconData get bannerIcon`.

witwash: Could you please elaborate on that? Why only `getBannerIcon`? So I'd spot cases like that by myself in the future.

ookami-kb: It turns out that we have the info on when to avoid getters, but not on when to use them :) https://github.com/MewsSystems/tech-mo/wiki/Mews-styleguide-for-Flutter-apps#getters-feel-faster-than-methods

In general, our undocumented (yet) style preference is to use the getter instead of `get*` method without parameters if it doesn't conflict with the above-mentioned style guide rule.

alboiuvlad29:
I feel like, for a notification widget, it should have some sort of animation. Fade in/out with some curves or some slide in/out from the corner. And maybe I'm missing the code somewhere, but a public method like `showOptimusNotification(context, builder)` should be made available, similarly to what flutter provides (`showDialog`, `showBottomSheet`, `showOptimusDialog`)

witwash:
Animations and stacking will be implemented in [RND-104514](https://mews.myjetbrains.com/youtrack/issue/RND-104514)

ookami-kb:
+1 for `showOptimusNotification`.

optimus/lib/src/notification.dart
witwash: OK, I get it. Thanks

witwash:
It's still WIP as showing a notification is tightly connected with follow-up issue RND-104514, and I don't see how I can implement one without another. I think I will close this one and will split it into several, smaller PRs. What is the recommended way of doing it?

ookami-kb:
@witwash you can do the notification widget only, and showcase it in storybook as a normal widget.

PR #205, opened 2022-05-19
feat(kiosk_mode): Track Kiosk mode changes on Android

Sevastyan:
#### Summary

`watchKioskMode` now works on Android too!

#### Testing steps

Test that you get a stream of values for kiosk mode on Android. You can run an example app to check this.

#### Follow-up issues

No.

#### Check during review

- Unused code removed.
- Build passing.

### DIFF
--- a/kiosk_mode/lib/kiosk_mode.dart
+++ b/kiosk_mode/lib/kiosk_mode.dart
@@ -1,9 +1,15 @@
 import 'dart:async';
 
+import 'package:flutter/foundation.dart';
 import 'package:flutter/services.dart';
 import 'package:stream_transform/stream_transform.dart';
 
 const _channel = MethodChannel('com.mews.kiosk_mode/kiosk_mode');
+
+/// Channel used to send "kiosk mode" updates from platform.
+///
+/// Implementing it on Android platform is useless, as this plugin already
+/// does it. See [watchKioskMode] for explanation.
 const _eventChannel = EventChannel('com.mews.kiosk_mode/kiosk_mode_stream');
 
 /// Corresponds to lock task / screen pinning mode on Android, and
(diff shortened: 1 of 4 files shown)


ookami-kb (approved):

kiosk_mode/lib/kiosk_mode.dart
```diff
@@ -45,13 +51,23 @@ Future<KioskMode> getKioskMode() => _channel
...
+  Duration androidQueryPeriod = const Duration(seconds: 5),
+}) =>
+    Stream.fromFuture(getKioskMode())
+        .merge(_getKioskModeStream(androidQueryPeriod));
+
+Stream<KioskMode> _getKioskModeStream(Duration androidQueryPeriod) {
+  switch (defaultTargetPlatform) {
+    case TargetPlatform.android:
+      return Stream<void>.periodic(androidQueryPeriod)
+          .asyncMap((_) => getKioskMode());
```
alboiuvlad29, line 66: This should be done at a platform level, not plugin. I suggest updating the android side to do this kind of query and only add values to the event channel if they change.

alboiuvlad29: Sorry, haven't noticed it got merged. Maybe an issue can be created for this, and perhaps migrating sometime to [federated plugins](https://docs.flutter.dev/development/packages-and-plugins/developing-packages#federated-plugins).

ookami-kb: Yeah, it actually makes more sense to put it on the native side. Yes, please create an issue.

Regarding federated plugins – I'm not sure if it makes sense in our case. AFAIK old way is not deprecated yet, and federated plugins bring more overhead.

alboiuvlad29: Yet :)
Yes it might bring a bit more overhead, but also would make it easier for different teams to work on it or someone might have moire expertise on android rather than iOS. And also we have the option to provide default implementation for unsupported platforms, without having to do the following (mind the pseudocode):
```
if platform.isAndroid
 // do android stuff
else if (platform.isIos) 
 // do iOS stuff
else
 // do default stuff
```
Either way this might not make that much sense in our case now,  but I'll try in an initiative in my free time to see how it would look like

Sevastyan: I don't think it is worth bothering with pushing it to native side. When it comes to cross-platform development, the less platform-specific code you have to write the better. We already know that Android doesn't support subscription to lock task mode changes, so we use what we already have - a simple method to get the state.

ookami-kb: The benefit of pushing it to the native side is that the platform channel won't be used until the value is actually changed.

alboiuvlad29: Also, currently the app mode change is triggered every 5 seconds, flooding the logs. Which is incorrect, since the mode didn't change.
image

PR #207, opened 2022-05-20
feat: kiosk_mode update [do not squash]

ookami-kb:


kiosk_mode/CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## 0.2.0
+
+> Note: This release has breaking changes.
+
+ - **FEAT**: Add `isManagedKiosk` method (#206).
+ - **FEAT**: Track Kiosk mode changes on Android (#205).
+ - **BREAKING** **FEAT**: Update readme.
```
Sevastyan, line 7: How is a README update a breaking change? Shouldn't it be for #206?

Sevastyan (approved):
Otherwise, LGTM.

kiosk_mode/CHANGELOG.md
ookami-kb: It should, but you didn't put it there and I missed it during review 😄

Sevastyan: But I did put it there 🤔
Screenshot_2022-05-20_17-15-46

ookami-kb: It's squashed, so the description isn't going into the final commit. It's better to use "!" in the commit message.

PR #210, opened 2022-05-25
feat(kiosk_mode): [RND-106334] move watchKioskMode implementation on platform side (android)

alboiuvlad29:
#### Summary

RND-106334

Based on [this discussion ](https://github.com/MewsSystems/mews-flutter/pull/205#discussion_r878016846), this PR moves the implementation of watchKioskMode periodic checks for android to the platform side.

The advantages are:
- no inconsistencies between android and iOS plugin, both use the EventChannel
- the stream will not emit the same value again if the mode hasn't changed
- Platform channel won't be used until the value actually changes

For better visibility, the android implementation has been extracted in functions. 

#### Testing steps

1. Listen to `watchKioskMode` stream on Android
2. Confirm that an event is not being sent every `androidQueryPeriod` like before, and only when kiosk mode actually changes
3. Confirm that `androidQueryPeriod` changes the checking period on the platform level

#### Follow-up issues

No.

#### Check during review

- Unused code removed.
- Build passing.

### DIFF
--- a/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt
+++ b/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt
@@ -12,24 +12,24 @@ import io.flutter.plugin.common.EventChannel
 import io.flutter.plugin.common.MethodCall
 import io.flutter.plugin.common.MethodChannel
 import io.flutter.plugin.common.MethodChannel.MethodCallHandler
-import io.flutter.plugin.common.MethodChannel.Result
 
 private const val methodChannelName = "com.mews.kiosk_mode/kiosk_mode"
-private const val kioskModeEventChannelName = "com.mews.kiosk_mode/kiosk_mode_stream"
+private const val eventChannelName = "com.mews.kiosk_mode/kiosk_mode_stream"
 
 class KioskModePlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
     private lateinit var channel: MethodChannel
+    private lateinit var eventChannel: EventChannel
     private var activity: Activity? = null
 
     override fun onAttachedToEngine(@NonNull flutterPluginBinding: FlutterPlugin.FlutterPluginBinding) {
         channel = MethodChannel(flutterPluginBinding.binaryMessenger, methodChannelName)
         channel.setMethodCallHandler(this)
 
-        EventChannel(flutterPluginBinding.binaryMessenger, kioskModeEventChannelName)
-            .setStreamHandler(KioskModeStreamHandler { isInKioskMode() })
+        eventChannel = EventChannel(flutterPluginBinding.binaryMessenger, eventChannelName)
+        eventChannel.setStreamHandler(KioskModeStreamHandler { isInKioskMode()})
     }
 
-    override fun onMethodCall(@NonNull call: MethodCall, @NonNull result: Result) {
+    override fun onMethodCall(@NonNull call: MethodCall, @NonNull result: MethodChannel.Result) {
         when (call.method) {
             "startKioskMode" -> startKioskMode(result)
             "stopKioskMode" -> stopKioskMode(result)
@@ -39,7 +39,7 @@ class KioskModePlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
         }
     }
 
-    private fun startKioskMode(@NonNull result: Result) {
+    private fun startKioskMode(@NonNull result: MethodChannel.Result) {
         activity?.let { a ->
             // ensures that startLockTask() will not throw
             // see https://stackoverflow.com/questions/27826431/activity-startlocktask-occasionally-throws-illegalargumentexception
@@ -54,13 +54,13 @@ class KioskModePlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
         } ?: result.success(false)
     }
 
-    private fun stopKioskMode(@NonNull result: Result) {
+    private fun stopKioskMode(@NonNull result: MethodChannel.Result) {
         activity?.stopLockTask()
         result.success(null)
     }
 
 
-    private fun isManagedKiosk(@NonNull result: Result) {
+    private fun isManagedKiosk(@NonNull result: MethodChannel.Result) {
         val service = activity?.getSystemService(Context.ACTIVITY_SERVICE) as? ActivityManager
         if (service == null) {
             result.success(null)
@@ -70,13 +70,13 @@ class KioskModePlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
         result.success(service.lockTaskModeState == ActivityManager.LOCK_TASK_MODE_LOCKED)
     }
 
-    private fun isInKioskMode(@NonNull result: Result) {
+    private fun isInKioskMode(@NonNull result: MethodChannel.Result) {
         result.success(isInKioskMode())
     }
 
-    private fun isInKioskMode(): Boolean? {
+    private fun isInKioskMode(): Boolean {
         val service = activity?.getSystemService(Context.ACTIVITY_SERVICE) as? ActivityManager
-            ?: return null
+            ?: return false
 
         val isInKioskMode = when (service.lockTaskModeState) {
             ActivityManager.LOCK_TASK_MODE_NONE -> false
@@ -90,6 +90,7 @@ class KioskModePlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
 
     override fun onDetachedFromEngine(@NonNull binding: FlutterPlugin.FlutterPluginBinding) {
         channel.setMethodCallHandler(null)
+        eventChannel.setStreamHandler(null)
     }
 
     override fun onAttachedToActivity(binding: ActivityPluginBinding) {
--- a/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt
+++ b/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt
@@ -4,7 +4,6 @@ import android.app.Activity
 import android.app.ActivityManager
 import android.content.Context
 import android.view.ViewGroup
-import androidx.annotation.NonNull
 import io.flutter.embedding.engine.plugins.FlutterPlugin
 import io.flutter.embedding.engine.plugins.activity.ActivityAware
 import io.flutter.embedding.engine.plugins.activity.ActivityPluginBinding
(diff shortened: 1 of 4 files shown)


kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModeStreamHandler.kt
```diff
@@ -0,0 +1,37 @@
...
+                previousIsKioskModeState = newIsKioskModeState
+                uiThreadHandler.post() {
+                    eventSink?.success(newIsKioskModeState)
+                }
+            }
+        }
+    }
+
+    override fun onCancel(arguments: Any?) {
+        eventSink = null
```
Sevastyan, line 33: also do
```kotlin
eventSink?.endOfStream()
```

kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModeStreamHandler.kt
```diff
@@ -0,0 +1,37 @@
+package com.mews.kiosk_mode
+
+import android.os.Handler
+import android.os.Looper
+import io.flutter.plugin.common.EventChannel
+import java.util.*
```
Sevastyan, line 6: Don't do this. Do this:

```
import java.util.Timer
```

kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModeStreamHandler.kt
```diff
@@ -0,0 +1,37 @@
...
+    private var timer: Timer? = null
+    private var previousIsKioskModeState: Boolean? = null
+    private val uiThreadHandler: Handler = Handler(Looper.getMainLooper())
+
+    override fun onListen(arguments: Any?, events: EventChannel.EventSink?) {
+        val period: Int = (arguments as Map<*, *>)["androidQueryPeriod"] as Int
+
+        eventSink = events
+        timer = fixedRateTimer(period = period.toLong(), daemon = true) {
+            val newIsKioskModeState = isKioskMode()
```
Sevastyan, line 22: I strongly suggest using

```
currentState
previousState
```

Reading `isKioskMode` in the name is just distracting.

kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModeStreamHandler.kt
```diff
@@ -0,0 +1,37 @@
+package com.mews.kiosk_mode
+
+import android.os.Handler
+import android.os.Looper
+import io.flutter.plugin.common.EventChannel
+import java.util.*
+import kotlin.concurrent.fixedRateTimer
+
+typealias KioskModeHandler = () -> Boolean?
```
Sevastyan, line 9: Don't use `typedef` and don't declare the return type as nullable.

kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModeStreamHandler.kt
```diff
@@ -0,0 +1,37 @@
...
+
+class KioskModeStreamHandler(val isKioskMode: KioskModeHandler) : EventChannel.StreamHandler {
+    private var eventSink: EventChannel.EventSink? = null
+    private var timer: Timer? = null
+    private var previousIsKioskModeState: Boolean? = null
+    private val uiThreadHandler: Handler = Handler(Looper.getMainLooper())
+
+    override fun onListen(arguments: Any?, events: EventChannel.EventSink?) {
+        val period: Int = (arguments as Map<*, *>)["androidQueryPeriod"] as Int
```
Sevastyan, line 18: Better to pass this value to the constructor of `KioskModeStreamHandler`. Otherwise, there is this knowledge spread about platform arguments that really should be kept to the `FlutterPlugin` only.

kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModeStreamHandler.kt
```diff
@@ -0,0 +1,37 @@
...
+
+        eventSink = events
+        timer = fixedRateTimer(period = period.toLong(), daemon = true) {
+            val newIsKioskModeState = isKioskMode()
+            if (newIsKioskModeState != previousIsKioskModeState) {
+                previousIsKioskModeState = newIsKioskModeState
+                uiThreadHandler.post() {
+                    eventSink?.success(newIsKioskModeState)
+                }
```
Sevastyan, line 27: Don't post anything to the UI thread here. Just call a callback function (to be declared in the constructor).

Sevastyan (changes requested):
This is not a `chore`, but  a fully-fledged `feat`.

kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModeStreamHandler.kt
alboiuvlad29: No way around it. A callback exposed to `KioskModePlugin` class will not really help. This is the point where a new value is sent to the stream, but this can only be accomplished on the UI thread since `setStreamHandler` is declared with`  @UiThread` annotation. This should be the logic of `KioskModeStreamHandler`.

kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModeStreamHandler.kt
alboiuvlad29: I do not agree with this. The parameter is not passed by a method channel, but by the event channel when the broadcast stream is started and listened to. All the logic is self contained within the `KioskModeStreamHandler`. While I am also not that happy with "magic" variables, potentially I could make this class constant.

Sevastyan (changes requested):

kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModeStreamHandler.kt
alboiuvlad29: > The parameter is passed to the platform no matter what kind of channel that is.

How is that? The parameter is just passed to this specific event channel, in this case with the name `com.mews.kiosk_mode/kiosk_mode_stream`. It doesn't make sense to send unrelated arguments to all channels. Then, from platform side `KioskModeStreamHandler` should be the only one handling anything related to this particular event channel. I don't see why we would have to mix this.

```dart
    _eventChannel.receiveBroadcastStream(
      {'androidQueryPeriod': androidQueryPeriod.inMilliseconds},
    ).
```

alboiuvlad29: As for your other comment, this is how other platforms do it:
https://github.com/fluttercommunity/plus_plugins/blob/e81d039d34cee4cb898962880c46c13de4278a83/packages/connectivity_plus/connectivity_plus/android/src/main/java/dev/fluttercommunity/plus/connectivity/ConnectivityBroadcastReceiver.java
 ```kotlin 
 Runnable runnable = () -> events.success(connectivity.getNetworkType());
 mainHandler.post(runnable);
```

**Review by Sevastyan — CHANGES_REQUESTED** [review]

File: kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModeStreamHandler.kt (hunk 5 above)
Comment (alboiuvlad29) [reply kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModeStreamHandler.kt:17]: I see your point. But for this case the plugin is specifically made for Android, so saying that it has nothing to do with Android seems a bit erroneous, as this would be the point where the native host would be notified of changes. I think just having `KioskModeStreamHandler` as a black box makes a bit more sense, as in "plug-it" and forget it approach. We just know it's a handler for a particular channel, so no need to also query for parameters there.
 @ookami-kb  What do you think?

File: kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt [hunk 7]
```diff
@@ -4,71 +4,92 @@ import android.app.Activity
...
+
+    private fun isInKioskMode(result: MethodChannel.Result) {
+        result.success(isInKioskMode())
+    }
+
+    private fun isInKioskMode(): Boolean? {
+        val service = activity?.getSystemService(Context.ACTIVITY_SERVICE) as? ActivityManager
+            ?: return null
+
+        val isInKioskMode = when (service.lockTaskModeState) {
```
ookami-kb, line 80: I would rather inline it.

kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModeStreamHandler.kt
ookami-kb: `getQueryPeriodMilliseconds` as a constructor parameter makes no sense to me, it makes the solution less readable, and I don't really see what it is needed for. To hide the knowledge that the actual parameter on Flutter side is called `androidQueryPeriod`? Well, the arguments are passed to the `onListen` method anyway, so it makes total sense to process them there, why should `KioskModePlugin` know about it?

Sevastyan:
@ookami-kb

> why should KioskModePlugin know about it?

It is better to keep knowledge of the parameters in one place. Especially in this case when we don't event have a constants for their names.

alboiuvlad29:
I agree more with @ookami-kb about this, and this is a pattern similarly seen in other packages. There is no need for argument parameter names to be kept in one place, in the plugin file. Other packages  have declared StreamHandlers within the plugin file, but to me this makes things less readable and should be in a separate file. Not sure what other argument to bring to the table but I'm keen to find out what @ookami-kb can say about this, and hopefully yo move this PR further.

ookami-kb:
> It is better to keep knowledge of the parameters in one place. Especially in this case when we don't event have a constants for their names.

The parameters are not related to the plugin in general. They are specific to watching the Kiosk mode, so it doesn't make sense for the whole plugin to know about them.

ookami-kb (approved):

Sevastyan (changes requested):

kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModeStreamHandler.kt
```diff
@@ -0,0 +1,36 @@
...
+
+
+class KioskModeStreamHandler(val getKioskModeState: () -> Boolean) : EventChannel.StreamHandler {
+    private var eventSink: EventChannel.EventSink? = null
+    private var timer: Timer? = null
+    private var previousState: Boolean = getKioskModeState()
+    private val mainHandler: Handler = Handler(Looper.getMainLooper())
+
+    override fun onListen(arguments: Any?, events: EventChannel.EventSink?) {
+        val period: Int = (arguments as Map<*, *>)["androidQueryPeriod"] as Int
```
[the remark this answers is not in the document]
alboiuvlad29: This line doesn't work, I tried it initially:
`arguments as Map<*, *>)["androidQueryPeriod"] as Long`. It throws saying you can't cast int to long.

Sevastyan (approved):

Sevastyan:
ping @ookami-kb

PR #211, opened 2022-05-26
feat(optimus)!: [RND-103047] Add Notification showing and stacking [ignore-size]

witwash:
#### Summary

RND-103047 Added showing and stacking notifications. 

<details>
<summary>Video</summary>

https://user-images.githubusercontent.com/9210422/178459960-c9e99b68-c268-4f34-94e9-c7419631beac.mp4

</details>

Breaking changes:

- `OptimusNotification` now requires `OptimusNotificationLink`, to assure both **link text** and **link callback** are provided. Or neither of them.

#### Testing steps

1. Open Storybook / Test in your project (wrap screen with OptimusNotificationsOverlay)
2. Try to show notifications inside this screen. 
3. Notifications should look and behave according to the Optimus Design System guidelines. 

#### Follow-up issues

- ~~RND-108205~~ fixed here.
- RND-114092 - fix code after a breaking change. 

#### Check during review

- Unused code removed.
- Build passing.


optimus/lib/src/notification/notification_manager.dart
```diff
@@ -0,0 +1,278 @@
...
+    });
+
+    return notification;
+  }
+
+  Widget _buildRemovedNotification(
+    _NotificationModel removedItem,
+    Animation<double> animation,
+  ) {
+    animation.addStatusListener(
```
witwash, line 212: I've discovered that this can cause AnimationController to be disposed more than once.
`Error: AnimationController.dispose() called more than once.
A given AnimationController cannot be disposed more than once.
The following AnimationController object was disposed multiple times:
  AnimationController#d9d78(⏮ 0.000; paused; DISPOSED)
    at Object.throw_ [as throw] (http://localhost:56084/dart_sdk.js:5080:11)
at http://localhost:56084/packages/flutter/src/animation/animation_controller.dart.lib.js:477:21
at animation_controller.AnimationController.new.dispose (http://localhost:56084/packages/flutter/src/animation/animation_controller.dart.lib.js:480:25)
at http://localhost:56084/packages/flutter/src/widgets/animated_list.dart.lib.js:561:124
    at _RootZone.runUnary (http://localhost:56084/dart_sdk.js:40511:59)
    at _FutureListener.then.handleValue (http://localhost:56084/dart_sdk.js:35438:29)
    at handleValueCallback (http://localhost:56084/dart_sdk.js:35999:49)
    at Function._propagateToListeners (http://localhost:56084/dart_sdk.js:36037:17)
    at _Future.new.[_completeWithValue] (http://localhost:56084/dart_sdk.js:35872:23)
    at async._AsyncCallbackEntry.new.callback (http://localhost:56084/dart_sdk.js:35906:35)
    at Object._microtaskLoop (http://localhost:56084/dart_sdk.js:40778:13)
    at _startMicrotaskLoop (http://localhost:56084/dart_sdk.js:40784:13)
    at http://localhost:56084/dart_sdk.js:36261:9`
If I remove this listener then the problem is gone, but I have no clue when removing animation is done and the last notification is removed with overlay immediately, without removing animation. Any suggestions?

alboiuvlad29: @witwash can you not something like this instead? I don't think it's the builder's job to do that. Also removeItem will remove the item straight away.
```diff
void _removeNotification(_NotificationModel model) {
      final index = widget._notifications.indexOf(model);
+     model.onNotificationDismissed.call()
      widget._notifications.removeAt(index);
      widget._listStateKey.currentState?.removeItem(
        index,
-        (context, animation) => _buildRemovedNotification(model, animation),
+       (context, animation) => _buildRemovedNotification(animation),
        duration: _animationDuration,
      );
    }
   
```

witwash: I had something like this, but then when I check if there are no notifications and I can remove the overlay I had to delay it by the animation time. Because otherwise last notification would be deleted immediately, before the animation end.

alboiuvlad29: Not sure wether this helps, but doesn't seem to be a framework issue.
image
ezgif-5-cc93bb1b6e

I think maybe some refactoring is in order there may be some concurrency issue somewhere. For example, this code is being recreated every ms during the animation period, in `_buildNotification`:
```dart
 Future<void>.delayed(_autoDismissDuration, () {
       if (widget._notifications.contains(model)) {
         _removeNotification(model);
       }
   }); 
  ```

Comment (witwash) [reply optimus/lib/src/notification/notification_manager.dart:212]: Yeah, good point, I'll look into that. The problem is that `removeAt(index)` is executed before `removeItem` animation ends. Therefore you are animating an item that is already removed. And in the case when I check if the list is empty, I have no idea if animation is in progress. The list is empty and I can remove the overlay. Thus removing overlay while the last item is animating its departure.

File: optimus/lib/src/notification/notification_manager.dart [hunk 2]
```diff
@@ -0,0 +1,278 @@
...
+  final String? link;
+  final OptimusNotificationVariant variant;
+  final VoidCallback? onLinkPressed;
+  final VoidCallback onNotificationDismissed;
+  final VoidCallback? onDismissed;
+}
+
+const Duration _animationDuration = Duration(milliseconds: 500);
+const Duration _autoDismissDuration = Duration(seconds: 8);
+const int _maxVisibleCount = 3;
```
Sevastyan, line 277: Please, make sure that an external client can define a limit of `1`, so the notification can be replaced. It is a use-case for Kiosk.

optimus/lib/src/notification/notification_manager.dart
witwash: I've added a slight delay after removing a now this issue is sorted out. Thanks for your help.

optimus/lib/src/notification/notification_manager.dart
witwash: I've added an option to set `customMaxVisibleCount` and change the maximum visible count. I'd like to have your opinion on what should be the initial value, 1 or 3?

witwash: @Sevastyan ping

Sevastyan: If the value can be overwritten, I really don't care  😄 

But if you need an answer, I'd say `3` is a sensible default.

@witwash

optimus/lib/src/notification/notification_manager.dart
```diff
@@ -0,0 +1,329 @@
...
+    required BuildContext context,
+    required String title,
+    String? body,
+    String? link,
+    IconData? icon,
+    VoidCallback? onLinkPressed,
+    VoidCallback? onDismissed,
+    OptimusNotificationVariant variant = OptimusNotificationVariant.info,
+  }) {
+    assert(
```
ookami-kb, line 38: Instead of doing an assert, create `NotificationLink` mode with non-nullable link and callback, in that case illegal state would become impossible.

optimus/lib/src/notification/notification_manager.dart
```diff
@@ -0,0 +1,329 @@
...
+    late _NotificationModel notification;
+    final onDismissPressed = onDismissed == null
+        ? null
+        : () {
+            _removeNotification(notification);
+          };
+    final onLinkPress = onLinkPressed == null
+        ? null
+        : () {
+            onLinkPressed.call();
```
ookami-kb, line 74:
```suggestion
            onLinkPressed();
```

Same in other places.

optimus/lib/src/notification/notification_manager.dart
```diff
@@ -0,0 +1,329 @@
...
+    required GlobalKey<AnimatedListState> listStateKey,
+    required List<_NotificationModel> notifications,
+    required VoidCallback onVisible,
+  })  : _listStateKey = listStateKey,
+        _notifications = notifications,
+        _onVisible = onVisible,
+        super(key: key);
+  final GlobalKey<AnimatedListState> _listStateKey;
+
+  final List<_NotificationModel> _notifications;
```
ookami-kb, line 259: No need to make widget's final properties private.

optimus/lib/src/notification/notification_manager.dart
```diff
@@ -0,0 +1,329 @@
...
+  /// otherwise first notification would be displayed without animation.
+  @override
+  void initState() {
+    super.initState();
+    WidgetsBinding.instance.addPostFrameCallback((_) {
+      widget._onVisible.call();
+    });
+  }
+
+  Widget _buildNotification(int index, Animation<double> animation) {
```
ookami-kb, line 277: Since it's called `_buildNotification` I would expect it to accept notification rather than index.

In general, methods should be as clean as possible – it's better when the method depends on the parameters rather than on inner state (it's not possible or makes sense everywhere, but it should be a preferable option by default).

ookami-kb (changes requested):

optimus/lib/src/notification/notification_manager.dart
```diff
@@ -0,0 +1,329 @@
...
+  }
+
+  OverlayEntry _buildOverlayEntry({
+    required _NotificationModel initialEntry,
+  }) =>
+      OverlayEntry(
+        builder: (context) => Positioned(
+          left: _getLeftPadding(context),
+          right: _getRightPadding(context),
+          child: SafeArea(
```
alboiuvlad29, line 134: Using safe area like this doesn't look good. For example, if running this on a device with notch (or even iPad with the home indicator area) will "cut off" notification on scrolling. From a UI perspective this doesn't look nice, and seeing apps that do this is on iOS  but even other platforms makes the app look unprofessional and not aligned with the native UI. I suggest Insetting padding accordingly to the list items instead:
- List item is first, insert top padding by `MediaQuery.of(context).padding.top`
- List item is last, insert bottom padding by `MediaQuery.of(context).padding.bottom`

Horizontal padding is OK to be defined by SafeArea, as there is not cut off.

alboiuvlad29 (changes requested):

witwash:
Apparently `SizeTransition` is cutting shadows for some reason, so I switched to `SlideTransition` and I'll replace it with a custom animated widget that will handle that jump when there are several notifications on the screen. 

https://user-images.githubusercontent.com/9210422/172631951-e0be415e-4ed2-4fa2-9dfc-17799cea7aaf.mp4

For now, I removed the possibility to change the **number of visible notifications** and it's now set to **1**. I'll create a follow-up issue with this problem and will fix it with the next PR.

alboiuvlad29:
Sounds good, please add a follow up issue in the description for this.
I've approved your PR, LGTM.

witwash:
Here is a demonstration with `SizeTransition`

https://user-images.githubusercontent.com/9210422/172636119-44d05f4a-330e-4e18-87cb-a883ab94b27f.mp4

witwash:
@ookami-kb can you please take a look at it?

optimus/lib/src/notification/notification_manager.dart
```diff
@@ -0,0 +1,349 @@
+import 'package:flutter/material.dart';
+import 'package:optimus/optimus.dart';
+
+/// The OptimusNotificationManager serves for handling OptimusNotifications.
+///
+/// By default maximum visible notifications is set to [_defaultMaxVisibleCount].
+class OptimusNotificationManager {
```
ookami-kb, line 7: Do we actually need an overlay here?

Probably, the easier solution would be to just go with `Stack`: e.g. create a `NotificationsOverlay` widget that will wrap the child in stack and add a column with notifications on top of it. It will also allow placing "notifications" anywhere in the app by wrapping the corresponding part of the app in `NotificationsOverlay`. WDYT?

witwash: I chose overlay because you never know where would you need to call notifications from, and ideally their state should be preserved across the application. Wrapping it all with `Stack` could be a good alternative solution, I'll look into that. Nevertheless if would require you to mind where do you put it. 

By design, notifications could be placed either in the top right corner or at the top center, but I'm all for being able to place them anywhere on the screen.

alboiuvlad29:
Hey @witwash if you'd like, you can use the transitions I've done for this PR:
https://github.com/MewsSystems/tech-mo/pull/1290. They seem snappier and smoother.

In particular you'd be looking for `_NotificationTransition`.

witwash:
@alboiuvlad29 yes, I was just thinking about that. It does, indeed, look smoother.

witwash:
@alboiuvlad29 can you please take a look at it again? 🙏

alboiuvlad29 (approved):

witwash:
@ookami-kb, can you please take a look at it?

ookami-kb (changes requested):
I've pushed some small code style changes.

A couple of things:

1. Regarding animation: for me dismissing animation looks weird, when e.g. notification appears from bottom and goes away to the bottom as well, overlapping with other notifications. Dismissing to the side would be more natural, but it's better to discuss with UX guys.
2. Can you probably use `reverse` animated list instead of determine some properties based on direction? I think it would slightly simplify the code.

witwash:
@ookami-kb ping

optimus/lib/src/notification/notification_layer.dart
```diff
@@ -0,0 +1,341 @@
...
+/// overlay. Notifications that could not be shown because of the [maxVisible]
+/// limit will be put into a queue and will be dispatched in the order they were
+/// added.
+class OptimusNotificationsOverlay extends StatefulWidget {
+  const OptimusNotificationsOverlay({
+    Key? key,
+    this.left,
+    this.top,
+    this.right,
+    this.bottom,
```
ookami-kb, line 26: If I don't specify these parameters, the default values look weird – it slides from the center to the left.

In general, I would say we don't need this level of customization. We're not building a common and customizable UI library, this is a very opinionated design system with many restrictions.

I'd suggest having just one parameter – something like `enum Position { topLeft, topRight, bottomLeft, bottomRight }`, and notification position, animation direction, and stacking duration should be calculated based on it.

ookami-kb (changes requested):

optimus/lib/src/notification/notification_layer.dart
witwash: How about a defined set of `Position`s with a default value, but you can change the stacking order and slide direction. 
```
inDirection: customInDirection ?? position.inDirection
stackDirection: customStackDirection ?? position.stackDirection 
```

witwash:
@ookami-kb ping

witwash:
@ookami-kb can you take a look at it again, please?

optimus/lib/src/notification/notification_layer.dart
```diff
@@ -0,0 +1,374 @@
...
+/// [OptimusNotificationsOverlay] is an overlay that wraps [child] inside a
+/// stack and provides a method to show an in-app notification.
+///
+/// The position of notifications is determined by [OptimusNotificationPosition]
+/// and by default is set to [OptimusNotificationPosition.topRight].
+/// You can change the maximum visible count [maxVisible] when declaring the
+/// overlay. Notifications that could not be shown because of the [maxVisible]
+/// limit will be put into a queue and will be dispatched in the order they were
+/// added.
+class OptimusNotificationsOverlay extends StatefulWidget {
```
ookami-kb, line 15: Let's call the file `notifcation_overlay` in that case.

optimus/lib/src/notification/notification_layer.dart
```diff
@@ -0,0 +1,374 @@
...
+            ),
+          ),
+        ],
+      ),
+    );
+  }
+}
+
+abstract class OptimusNotificationManager {
+  void show({
```
ookami-kb, line 175: For me, it would make more sense if this method would accept the `Notification` itself, rather than accepting the options and building a notification internally.

ookami-kb (changes requested):
Let's maybe have a call about it.

witwash:
I've updated the description. 

**New part:**
Breaking changes:

`OptimusNotification` now requires `OptimusNotificationLink`, to assure both link text and link callback are provided. Or neither of them.

@ookami-kb take a look, please 🙏

optimus/lib/src/notification/notification_overlay.dart
```diff
@@ -0,0 +1,325 @@
...
+
+  @override
+  void remove(OptimusNotification notification) {
+    final _NotificationModel? model = _findNotification(notification);
+    if (model != null) {
+      _removeNotification(model);
+    }
+  }
+
+  _NotificationModel? _findNotification(OptimusNotification notification) {
```
witwash, line 64: I'll replace this part with `firstOrNull`, when new `dfunc` will be out.

optimus/lib/src/notification/notification_overlay.dart
```diff
@@ -0,0 +1,325 @@
...
+      case OptimusNotificationPosition.topRight:
+        return false;
+      case OptimusNotificationPosition.bottomLeft:
+      case OptimusNotificationPosition.bottomRight:
+        return true;
+    }
+  }
+}
+
+class _NotificationModel {
```
ookami-kb, line 312: This class can be removed, you can use `OptimusNotification` directly.

ookami-kb (changes requested):

optimus/lib/src/notification/notification_overlay.dart
ookami-kb: Published

witwash:
@ookami-kb done

ookami-kb (approved):

PR #212, opened 2022-05-26
fix(optimus): [RND-103444] Fix visual issues in StepBar components

witwash:
#### Summary
Fixed separator not changing color with completed steps. Issue mentions numbers not being aligned inside the step widget, but I don't see it in `master`. Could be some inconsistency with https://mews-optimus.web.app/#/stories/step-bar.

RND-103444
image

#### Testing steps

Open Storybook/Step Bar and fiddle with it. The separator's color should change, depending on the current step state.

#### Follow-up issues

None

#### Check during review

- Unused code removed.
- Build passing.


chaeMil (approved):

optimus/lib/src/step_bar.dart
```diff
@@ -84,14 +84,32 @@ class _OptimusStepBarState extends State<OptimusStepBar> with ThemeGetter {
...
-            child: Container(color: theme.colors.primary),
+            child: Container(color: color),
           ),
         );
     }
   }
 
-  List<Widget> _buildItems(List<OptimusStepBarItem> items, double maxWidth) =>
-      items.map((i) => _buildItem(i, maxWidth)).intersperse(_spacer).toList();
+  List<Widget> _buildItems(List<OptimusStepBarItem> items, double maxWidth) {
```
ookami-kb, line 93: Why don't you use `intersperse` here?

ookami-kb: Let's maybe update the `intersperse` signature and make it more generic:

```dart
typedef IntersperseItemBuilder = T Function(T current);

Iterable<T> intersperse(IntersperseItemBuilder build) sync*
```

witwash: @ookami-kb I was trying to come up with some way to solve it using `intersperse`, but I don't think (or maybe don't see how) changing the signature and making it more generic would it solve my problem. We're interspersing Widgets with some other Widget. But in that case, I need the `OptimusStepBarItem`, not the Widget that was built using it, so I can build Spacer.

ookami-kb: I've added `intersperseWith` method to dfunc: https://pub.dev/documentation/dfunc/latest/dfunc/Intersperse/intersperseWith.html

You can use it. There's also `intersperse` method, so you can replace its usages and remove `intersperse` from utils.

optimus/lib/src/step_bar.dart
```diff
@@ -64,14 +64,16 @@ class _OptimusStepBarState extends State<OptimusStepBar> with ThemeGetter {
     return OptimusStepBarItemState.disabled;
   }
 
-  Widget get _spacer {
+  Widget _spacer(OptimusStepBarItem item) {
```
ookami-kb, line 67: As it's a method now, rename it to `_buildSpacer`.

ookami-kb (commented):
One small point, otherwise LGTM.

optimus/lib/src/step_bar.dart
witwash: Yes, good point. Done.

ookami-kb (approved):

PR #223, opened 2022-06-20
feat(optimus): [RND-103475] Added Checkbox nested variant

witwash:
#### Summary

Added nested checkboxes to Optimus. 

<details>
<summary>Screenshot</summary>

CleanShot 2022-06-20 at 11 20 02

</details>

#### Testing steps

- Open the nested checkboxes story and test it the parent checkbox represents the group state correctly.
	-  Checked - all children are checked.
	-  Unchecked - none of the children is checked.
	-  Dash - some children are checked.  

#### Follow-up issues

None

#### Check during review

- Unused code removed.
- Build passing.


witwash:
Bump

witwash:
@alboiuvlad29 can you please take a look?

alboiuvlad29 (changes requested):
From my point of view, I think it would make more sense to map `isChecked` and `tristate` to some internal state. I.e., 
```dart
enum _CheckboxState {checked, unchecked, undetermined }
```
This would be easier to handle in code.

optimus/lib/src/checkbox/checkbox.dart
```diff
@@ -107,18 +121,33 @@ class _OptimusCheckboxState extends State<OptimusCheckbox> with ThemeGetter {
...
+          ? Center(
+              child: Icon(
+                OptimusIcons.done,
+                size: 10,
+                color: theme.colors.neutral0,
+              ),
+            )
+          : null;
+    }
+  }
```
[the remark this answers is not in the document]
witwash: Sure thing, I was taking the inspiration from the origin tho :-/ : https://github.com/flutter/flutter/blob/676cefaaff197f27424942307668886253e1ec35/packages/flutter/lib/src/material/checkbox.dart#L212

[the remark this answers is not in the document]
alboiuvlad29: They are using MaterialState. It's ok to pass tristate as parameter, but use another internal state that "checked" and 'tristate' can be mapped to.

witwash:
@alboiuvlad29 ping 👀

optimus/lib/src/checkbox/checkbox.dart
```diff
@@ -107,18 +121,31 @@ class _OptimusCheckboxState extends State<OptimusCheckbox> with ThemeGetter {
...
         ? Center(
             child: Icon(
-              OptimusIcons.done,
+              _checkboxState.icon,
               size: 10,
               color: theme.colors.neutral0,
             ),
           )
         : null;
+  }
```
alboiuvlad29, line 136: This is fine, but prefer using Widgets rather than function builders when necessary in the future. And I think this is something we should emphasise in our docs, as I see we we rely in quite a couple of places on unnecessary builder methods. 
```dart
class _CheckboxIcon extends StatelessWidget {
const _CheckboxIcon({this.icon);

final IconData? icon;

@override
Widget build(BuildContext context) => icon == null ? const SizedBox() : Center(
             child: Icon(
               _checkboxState.icon,
               size: 10,
               color: theme.colors.neutral0,
             ),
           );
}
```

alboiuvlad29 (approved):

alboiuvlad29 (approved):

ookami-kb:
I meant here a different setup, from the user point of view you should be able to change checkboxes into a nested group by simply wrapping them into `OptimusCheckboxGroup`, smth like this:

```dart
OptimusCheckboxGroup(
  children: [
    OptimusCheckbox(),
    OptimusCheckbox(),
    OptimusCheckbox(),
  ],
)
```

Everything else should be done automatically.

witwash:
@ookami-kb I was trying to come up with the way I could do this with the way you mentioned. Each time I was stopped because I needed more control over the `OptimusCheckbox`es, therefore they need to be created inside the widget from some kind of model and not passed as a ready `Widget`. Can you please elaborate on the way you see it?

ookami-kb:
@witwash let's have a call about it.

witwash:
@ookami-kb can you please take a look?

witwash:
@ookami-kb after our call I made some changes. Now it's a mere wrapper that will change all the children inside it and will call their `onChange`, if the `parent` changes.

optimus/lib/src/checkbox/nested_checkbox.dart
```diff
@@ -0,0 +1,88 @@
...
+        label: widget.label,
+        error: widget.error,
+        child: OptimusEnabled(
+          isEnabled: widget.isEnabled,
+          child: Column(
+            mainAxisSize: MainAxisSize.min,
+            crossAxisAlignment: CrossAxisAlignment.start,
+            children: [
+              OptimusCheckbox(
+                isEnabled: widget.isEnabled,
```
ookami-kb, line 66: Not needed, you already wrap the whole group into `OptimusEnabled`.

optimus/lib/src/checkbox/nested_checkbox.dart
```diff
@@ -0,0 +1,88 @@
...
+                tristate: true,
+                isChecked: _isParentChecked,
+                label: widget.parent,
+                onChanged: (bool value) {
+                  for (final child in widget.children) {
+                    child.onChanged.call(value);
```
ookami-kb, line 72:
```dart
child.onChanged(value);
```

ookami-kb (changes requested):
One more thing: right now refreshing the group relies on the fact that changing the value of the single checkbox will call `setState` of the whole group widget thus leading to rebuild. I would say it should be more explicit – changing checkbox should explicitly invalidate the state of the group.

witwash:
@ookami-kb can you look at it again, please?

ookami-kb (approved):

PR #225, opened 2022-06-28
fix(optimus): [RND-73262] Misalignment of the text with radio button in Switch property option in both Andriod and iOS

witwash:
#### Summary

RND-73262

The whole row was aligned to the top, so the text widget wasn't aligned properly when the text could fit into one line.

<details>
<summary>Before</summary>

CleanShot 2022-06-28 at 18 58 06@2x

</details>

<details>

<summary>After</summary>

CleanShot 2022-06-28 at 18 57 32@2x

</details>

#### Testing steps

1. Open Storybook
2. Check Radio Story if the OptimusRadio displayed correctly.

#### Follow-up issues

None.

#### Check during review

- Unused code removed.
- Build passing.

### DIFF
--- a/optimus/lib/src/radio/radio.dart
+++ b/optimus/lib/src/radio/radio.dart
@@ -119,23 +119,28 @@ class _OptimusRadioState<T> extends State<OptimusRadio<T>> with ThemeGetter {
               onTapDown: (_) => setState(() => _isTappedDown = true),
               onTapUp: (_) => setState(() => _isTappedDown = false),
               onTapCancel: () => setState(() => _isTappedDown = false),
-              child: Row(
-                crossAxisAlignment: CrossAxisAlignment.start,
-                children: [
-                  _RadioCircle(
-                    isSelected: _isSelected,
-                    isActive: _isHovering || _isTappedDown,
-                  ),
-                  Expanded(
-                    child: Padding(
-                      padding: const EdgeInsets.symmetric(vertical: 2),
-                      child: DefaultTextStyle.merge(
-                        style: _labelStyle,
-                        child: widget.label,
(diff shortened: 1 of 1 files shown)


optimus/lib/src/radio/radio.dart
```diff
@@ -119,23 +119,28 @@ class _OptimusRadioState<T> extends State<OptimusRadio<T>> with ThemeGetter {
...
-                    isSelected: _isSelected,
-                    isActive: _isHovering || _isTappedDown,
-                  ),
-                  Expanded(
-                    child: Padding(
-                      padding: const EdgeInsets.symmetric(vertical: 2),
-                      child: DefaultTextStyle.merge(
-                        style: _labelStyle,
-                        child: widget.label,
+              child: IntrinsicHeight(
```
alboiuvlad29, line 122: `IntrinsicHeight` should used sparsely, in exceptional cases. Due to how the layout is done in multiple passes, this widget is very expensive, with the worst case being O(n^2). I would try using an alternative. Why not `crossAxisAlignment: CrossAxisAlignment.center,`?

alboiuvlad29 (changes requested):

optimus/lib/src/radio/radio.dart
witwash: Because by design it should be aligned to the top, if there are several lines of text. Otherwise, it should be centered. 

CleanShot 2022-06-29 at 12 22 10

witwash:
@alboiuvlad29 can you please take a look at it again? 🙏

alboiuvlad29 (approved):
Much better! 👍🏻

ookami-kb (approved):

PR #228, opened 2022-07-05
feat(optimus)!: [RND-103473] Update Input component

witwash:
#### Summary

RND-103473

Updated Input component to match mews.design and the design of the web [component](https://apps.mews.com/mews-ui-storybook/release/0.748.2/index.html?path=/story/form-input-inputfield--default). 

Breaking changes:
- By design `prefix` and `suffix` should be just text widgets (with applied text style), that are displayed closest to the text input on both sides.
- Previous `prefix` and `suffix` are now `leadingWidget` and `tailingWidget`. They can be used for displaying a leading icon or leading image as well as for displaying a tailing icon or tailing widget, as is declared on mews.design.

Visually these changes will not break anything, but some changes are required so the usage is consistent across all products.

<details> <summary>Screenshots</summary>

CleanShot 2022-07-05 at 16 01 33@2x
CleanShot 2022-07-05 at 16 01 21@2x
CleanShot 2022-07-05 at 16 01 08@2x
CleanShot 2022-07-05 at 16 01 15@2x

</details>

Other changes:
- Focused state has a priority over the error state (the focused field is always colored as focused even if there is an error)
- Password show/hide input is a variation of the tailing widget and if enabled should replace it.
- The loading icon is neither a suffix text nor a tailing widget, so it should not replace them.
- Clear button is now visible when activated and the input field has non-empty input.
- Updated Storybook. 

#### Testing steps

1. Open Storybook
2. OptimusInputField component should match the screenshots above (position of widgets inside the component).

#### Follow-up issues

RND-103470 - update select input component (plus dropdown position seems to be off)
RND-111259 - update usage in code.

#### Check during review

- Unused code removed.
- Build passing.


optimus/lib/src/chat/optimus_chat_input.dart
```diff
@@ -33,7 +33,7 @@ class _OptimusChatInputState extends State<OptimusChatInput> {
         controller: _controller,
         maxLines: 4,
         minLines: 1,
-        suffix: OptimusEnabled(
+        tailingWidget: OptimusEnabled(
```
alboiuvlad29, line 36: Should it not be `trailing` ? 
I also think we should remove `Widget`, seems unnecessary.

optimus/lib/src/common/field_wrapper.dart
```diff
@@ -159,18 +157,19 @@ class _FieldWrapper extends State<FieldWrapper> with ThemeGetter {
           ? theme.colors.neutral0t64
           : theme.colors.neutral1000t64;
 
+  Widget _getStyled(Widget widget) => DefaultTextStyle.merge(
+        style: preset100s.copyWith(color: theme.colors.neutral1000t32),
+        child: _Icon(child: widget),
+      );
+
```
alboiuvlad29, line 164: Extract to widget

optimus/lib/src/form/input_form_field.dart
```diff
@@ -25,8 +25,10 @@ class OptimusInputFormField extends FormField<String> {
     bool autocorrect = true,
     bool hasBorders = true,
     bool isRequired = false,
+    Widget? leadingWidget,
```
alboiuvlad29, line 28: We should remove `Widget` from name. `leading` would be enough.

optimus/lib/src/input_field.dart
```diff
@@ -3,6 +3,19 @@ import 'package:flutter/services.dart';
...
+/// Most commonly found in form layouts as one of the most important form
+/// elements. Has a strict order of widgets:
+/// [leadingWidget] - leading icon or image. Should be non-interactive.
+/// [prefix] - styled text, displayed before the user input.
+/// <input> - field with the user input.
+/// [suffix] - styled text, displayed after the user input.
+/// <clearAllButton> - if [isClearEnabled] is true there will be a button for
+/// clearing the input.
+/// <passwordButton> or [tailingWidget] - tailing image or icon that has some
+/// interactive functionality. If [isPasswordField] is true, passwordButton will
```
alboiuvlad29, line 17: This comment should be added to the field `trailing`. It makes a bit more sense to warn the developers at that point, that their widget may be overridden by password button. Same with the other parameters. Doesn't seem that necessary to mention their usage in the class docs.

optimus/lib/src/input_field.dart
```diff
@@ -131,31 +153,74 @@ class _OptimusInputFieldState extends State<OptimusInputField>
...
   }
 
-  void _onSuffixTap() {
+  void _onClearAllTap() {
     _effectiveController.clear();
     widget.onChanged?.call('');
   }
 
-  Widget get _suffix {
+  bool get _isShowClearAll =>
```
alboiuvlad29, line 173: `_shouldShowClearAllButton`

optimus/lib/src/input_field.dart
```diff
@@ -131,31 +153,74 @@ class _OptimusInputFieldState extends State<OptimusInputField>
...
+
+  Widget get _passwordButton => GestureDetector(
+        onTap: () => setState(() {
+          _isShowPasswordEnabled = !_isShowPasswordEnabled;
+        }),
+        child: Icon(
+          _isShowPasswordEnabled ? OptimusIcons.hide : OptimusIcons.show,
+        ),
+      );
+
```
alboiuvlad29, line 184: Extract to widget

optimus/lib/src/input_field.dart
```diff
@@ -131,31 +153,74 @@ class _OptimusInputFieldState extends State<OptimusInputField>
...
+    return OptimusStack(
+      direction: Axis.horizontal,
+      spacing: OptimusStackSpacing.spacing100,
+      mainAxisSize: MainAxisSize.min,
+      children: [
+        if (leadingWidget != null) leadingWidget,
+        if (prefix != null) prefix,
+      ],
+    );
+  }
```
alboiuvlad29, line 202: Extract to widget. You could use a `const SizedBox()` rather than returning null.

optimus/lib/src/input_field.dart
```diff
@@ -131,31 +153,74 @@ class _OptimusInputFieldState extends State<OptimusInputField>
...
+        if (suffix != null) suffix,
+        if (_isShowClearAll) _ClearAllButton(onTap: _onClearAllTap),
+        if (widget.showLoader) _loader,
+        if (widget.isPasswordField)
+          _passwordButton
+        else if (tailingWidget != null)
+          tailingWidget
       ],
     );
   }
```
alboiuvlad29, line 226: Extract to widget.

optimus/lib/src/input_field.dart
```diff
@@ -211,28 +276,17 @@ class _OptimusInputFieldState extends State<OptimusInputField>
...
-            ),
         ],
       );
 
   void _onFocusChanged() {
     setState(() {});
   }
 
+  void _onTextChanged() {
+    setState(() {});
```
alboiuvlad29, line 287: Why is this necesarry?

alboiuvlad29 (changes requested):

optimus/lib/src/input_field.dart
witwash: It's deliberate. It has to be null, otherwise, `FieldWrapper` would count with it and will add padding.

optimus/lib/src/input_field.dart
witwash: Removed the second one and reused it for both controllers. Overall it's here for making sure `ClearAllButton` is visible if the input is not empty.

ookami-kb:
According to mews.design, we just have left detail and right detail:

image

What about having just `leading` and `trailing` parameters accepting `Widget[]` and applying default text styles?

witwash:
Having just `leading` and `trailing` parameters is good, but not enough. @tomhirsch emphasised on icons/images and prefixes having a strict order. `[<leading icon/image>, <prefix>, user input, <suffix>, <clear all>, <loader>, <trailing icon/image>]`, like it's implemented for the [web](https://apps.mews.com/mews-ui-storybook/release/0.744.1/index.html?path=/story/form-input-inputfield--default&knob-Caption=Caption&knob-Caption%20icon=integration_partners&knob-Clearable=true&knob-Has%20show%20password%20toggle=true&knob-Helper=Caption%20text,%20description%20or%20notification&knob-Indicative%20icon=search&knob-Interactive%20icon=visibility&knob-Label=Label&knob-Loading=true&knob-Placeholder=Placeholder&knob-Prefix=prefx&knob-Required=true&knob-Size=medium&knob-Suffix=sfx&knob-Variant=default&knob-size=medium).

CleanShot 2022-07-19 at 12 42 10@2x


</details>

witwash:
@ookami-kb ping

ookami-kb (approved):

PR #242, opened 2022-08-02
feat(optimus): [RND-103479] Add alignment prop to typography components

witwash:
#### Summary

- Added missing `align` property to the typography components. By default, it will resolve to the `TextAlign.left`, so it should not bring any visual changes as previously it was set to `TextAlign.start` anyway.
- Updated stories

`OptimusTypographyAlignment.left`

CleanShot 2022-08-02 at 07 26 51@2x

`OptimusTypographyAlignment.right`

CleanShot 2022-08-02 at 07 26 58@2x

`OptimusTypographyAlignment.center`

CleanShot 2022-08-02 at 07 27 09@2x

</details> 

#### Testing steps

1. Open storybook
2. Open any typography coponent
3. Change it to very long text
4. Change the `align` property and check if it is displaying correctly.

#### Follow-up issues

None.

#### Check during review

- Unused code removed.
- Build passing.


optimus/lib/src/typography/alignment.dart
```diff
@@ -0,0 +1,16 @@
...
+    switch (this) {
+      case OptimusTypographyAlignment.left:
+        return TextAlign.left;
+      case OptimusTypographyAlignment.right:
+        return TextAlign.right;
+      case OptimusTypographyAlignment.center:
+        return TextAlign.center;
+    }
+  }
+}
```
alboiuvlad29, line 16: Is this wrapper really necessary ? Why not use `TextAlign` as is?

alboiuvlad29 (changes requested):

optimus/lib/src/typography/alignment.dart
witwash: Because the design is defined for the `left`, `center` and `right`. `TextAlign` has `start` and `end` on top of that.

witwash: @alboiuvlad29 do you think it's better to skip the wrapper and simply pass the `TextAlign`? I did it this way because for example the `OptimusStack` there is an `OptimusStackAlignment`, which is doing pretty much the same thing and the idea is to limit the choice to the ones that are allowed. 

```
extension on OptimusStackAlignment {
  CrossAxisAlignment get crossAxisAlignment {
    switch (this) {
      case OptimusStackAlignment.start:
        return CrossAxisAlignment.start;
      case OptimusStackAlignment.center:
        return CrossAxisAlignment.center;
      case OptimusStackAlignment.end:
        return CrossAxisAlignment.end;
    }
  }

  MainAxisAlignment get mainAxisAlignment {
    switch (this) {
      case OptimusStackAlignment.start:
        return MainAxisAlignment.start;
      case OptimusStackAlignment.center:
        return MainAxisAlignment.center;
      case OptimusStackAlignment.end:
        return MainAxisAlignment.end;
    }
  }
}
```

alboiuvlad29: For me, it makes more sense to be universal and use types that flutter provides, for easier swapping, but depends what others think. @ookami-kb thoughts?

witwash:
@ookami-kb can you please look at it?

optimus/lib/src/typography/alignment.dart
ookami-kb: I agree with @alboiuvlad29, let's rather use `TextAlign`.

The main reason for `OptimusStackAlignment`, IIRC, was for the user not to care about `CrossAxisAlignment` vs `MainAxisAlignment`.

witwash: Done ;)

optimus/lib/src/typography/highlight.dart
```diff
@@ -10,17 +10,26 @@ import 'package:optimus/src/typography/typography.dart';
...
   /// The content of the highlight.
   ///
   /// Typically a [Text] widget.
   final Widget child;
 
+  /// The alignment of the highlight.
+  ///
+  /// Defaults to [TextAlign.left].
+  final TextAlign align;
+
```
alboiuvlad29, line 28: Use macros for repeated documentation:
https://pub.dev/packages/dartdoc#macros

The `The alignment of the highlight.` should be generalised to something like `The alignment of the text within this widget` or similar.

You can have something like:
```dart
/// {@template optimus.components.text.textAlignment}
///  The alignment of the text within this widget.
///
/// Defaults to [TextAlign.left].
/// {@endtemplate}
```

and use it:
```diff
- /// The alignment of the highlight.
- ///
- /// Defaults to [TextAlign.left].
+ /// {@macro optimus.components.text.textAlignment}
 final TextAlign align;
```

alboiuvlad29 (changes requested):

alboiuvlad29 (approved):

ookami-kb (approved):

ookami-kb (changes requested):
Shouldn't this `align` property be nullable? So that by default, we don't override this property, and it will be taken from the upper text style?

witwash:
@ookami-kb done ;)

optimus/lib/src/typography/caption.dart
```diff
@@ -37,7 +37,7 @@ class OptimusCaption extends StatelessWidget {
   ///
   /// Defaults to [TextAlign.left].
```
alboiuvlad29, line 38: @witwash Please update this. The default provided by flutter in Text widget is `TextAlign.start`, but maybe we shouldn't mention this here since it may change with the framework?

witwash: @alboiuvlad29 

The documentation is correct, `align` will be set to `TextAlign.left` if no value is given.

Yes, the default of the `Text` widget is `TextAlign.start`, but `OptimusTypography` default is `TextAlign.left`. `start` and `end` are there for the RTL languages and we don't support those yet anyway

ookami-kb: But Flutter supports them, I think we should rather default to `TextAlign.start`. For LTR languages that would make no difference, but for RTL languages it will be automatically supported.

ookami-kb (approved):

PR #246, opened 2022-08-12
feat(optimus): [RND-114990] Create DateInput component [ignore-size]

witwash:
#### Summary

RND-114990

Added `OptimusDateInputFormField` component.

- Added `OptimusDateInputFormField`, which will provide a way for the user to enter the date using the keyboard.
    - `OptimusDateInputFormField` will format the input according to the provided `placeholder` mask.
    - Return value on `onSubmit` is `DateTime?`
    - `placeholder` will be displayed as a hint for the whole time, until the date is entered completely.
    - The user input will be highlighted for better distinction from the placeholder.
-  Added story to display the component.

<details><summary>Video</summary>

https://user-images.githubusercontent.com/9210422/184345066-f8fa2504-225a-4015-b5b5-4a5cddf93b00.mp4
</details> 

#### Testing steps

1. Open Storybook
2. Open Date Input story
3. Try entering something:
    - knobs and component properties are working as intended

#### Follow-up issues

None.

#### Check during review

- Unused code removed.
- Build passing.


witwash:
@alboiuvlad29, can you please look at it?

optimus/lib/src/form/date_input.dart
```diff
@@ -0,0 +1,252 @@
...
+  ///
+  /// It will be displayed as a hint, until the whole date is entered.
+  final String placeholder;
+
+  /// Separator between the date parts.
+  final String separator;
+
+  final FormFieldSetter<String> onSaved;
+  final FormFieldValidator<String> validator;
+  final ValueChanged<String>? onSubmitted;
```
alboiuvlad29, line 39: Should be `ValueChanged<DateTime?>` and `FormFieldSetter<DateTime?>`. Since it's a DateTime field, it doesn't make sense to pass a `String` and pass the parsing to the user of the library. Seems like unnecessary repeated work that can be offset to this library. Since dart supports DateTime natively as well, we should use this instead.

Also, my reasoning why it's nullable, it would be from the value being removed from the field, and DateTime parser would be unable to parse the date.

optimus/lib/src/form/date_input.dart
```diff
@@ -0,0 +1,252 @@
...
+    this.onSubmitted,
+    this.label,
+    this.error,
+    this.caption,
+    this.secondaryCaption,
+    this.keyboardType,
+    this.autovalidateMode,
+    this.size = OptimusWidgetSize.large,
+    this.isEnabled = true,
+    this.placeholder = 'DD/MM/YYYY',
```
alboiuvlad29, line 24: This should be not hard coded, but instead, use something like `DateFormat.yMMMd()` instead to support different locales.

alboiuvlad29 (changes requested):

optimus/lib/src/form/date_input.dart
witwash: It does support different locales, you can change it while creating this component. 

The thing is that I don't have `intl` under my sleeve and that's why parsing is managed outside this package and that's why date format is a hardcoded string.

ookami-kb: I would say let's rather add `intl` dependency. In the end, it's from `dart.dev`.

Re-inventing the wheel + making it very limited in the supported format doesn't make sense for me here.

witwash: OK, I'll make it around intl

alboiuvlad29: As this library is public, it should be internationalised/localised whenever posible. It shouldn't be "Kiosk" or "Commander" agnostic, even though we use it mostly internally. Not supporting this it's a bit of an anti pattern, and if it were for us to just use this library internally, offsetting the locale pattern to the user of this library isn't the best approach.

optimus/lib/src/form/date_input.dart
```diff
@@ -0,0 +1,291 @@
...
+    this.caption,
+    this.secondaryCaption,
+    this.keyboardType,
+    this.size = OptimusWidgetSize.large,
+    this.isEnabled = true,
+    this.format = OptimusDateInputFormat.ddmmyyyy,
+    this.isClearEnabled = false,
+  }) : super(key: key);
+
+  /// Format of the date. It will define the placeholder and the separator.
```
Sevastyan, line 27: Separator is the same for any format.

optimus/lib/src/form/date_input.dart
```diff
@@ -0,0 +1,291 @@
...
+      onNonMatch: (String value) {
+        children.add(
+          TextSpan(
+            style: defaultStyle,
+            text: value,
+          ),
+        );
+
+        return value;
+      },
```
Sevastyan, line 134: NItpick: some commas may be deleted to save space and aid  readability:
```dart
  @override
  TextSpan buildTextSpan({required BuildContext context, TextStyle? style, required bool withComposing}) {
    final List<TextSpan> children = [];

    text.splitMapJoin(
      pattern,
      onMatch: (Match m) {
        children.add(TextSpan(style: matchStyle, text: m[0]));

        return m[0] ?? '';
      },
      onNonMatch: (String value) {
        children.add(TextSpan(style: defaultStyle, text: value));

        return value;
      },
    );

    return TextSpan(children: children, style: style);
  }
```

Sevastyan (changes requested):
Shouldn't functionality of a `caption` be implemented? The thing that says "Use format..." and that spells actual date entered?

Why not using monospaced font for the input and placeholder to avoid it jumping around?

Component should allow a client to alter placeholder's representation. For example, to set `ГГГГ` instead of `YYYY`.

optimus/lib/src/form/date_input.dart
alboiuvlad29: @Sevastyan Not true. In other countries, `.`(dot) is used to write the date:
18.08.2022(today's date). But even so, this would be handled by intl.

Sevastyan: @alboiuvlad29 I was talking about the code.

witwash:
@ookami-kb, can you please take a look at it?

alboiuvlad29:
@witwash There is a bugged behaviour, when the format is changed during typing. See this video:

https://user-images.githubusercontent.com/45460964/186527702-ee895570-de8f-4786-b468-536e5341b3fa.mov


You can see that it stops allowing you to type and the formatting gets messed up.

This Is the snipped of code in Kiosk.

```dart
 final translatedLabel = context.l10n.translate(widget.label);
    final initialDate = widget.value ?? widget.initialDate;

    return OptimusDateInputFormField(
      initialValue: initialDate?.toDateTime(),
      format: DateFormat.yMd(context.read<LocalizationBloc>().state.currentLanguageCode.languageCode),
      label: widget.isRequired ? '$translatedLabel *' : translatedLabel,
      error: widget.error,
      onSubmitted: (date) {
        if (date != null) {
          widget.onChanged(Date.fromDateTime(date));
        }
      },
    );

```

alboiuvlad29:
@witwash Thank you for the update! The bug no.3 above hasn't been fixed

https://user-images.githubusercontent.com/45460964/186685994-ff706dbc-9a75-48d4-be57-a6d7b8dd5022.mov

- So when deleting the last digit, then pressing delete again, you cannot insert any more digits.
- When removing the digits between separators, you have to press delete key 2 times.
Let `|` be the cursor .

For: 12/01`|`:
Press delete  → 12/0`|`
Press delete  → 12/`|`
Press delete  → 12`|`
Press delete  → 1`|`
Press delete  → `|`

While it should be:
Press delete  → 12/0`|`
Press delete  → 12/`|`
Press delete  → 1`|`
Press delete  → `|`

witwash:
@alboiuvlad29 ,

I've updated the way selection works after deleting the `date` symbol around the `mask` symbol. 
I can't reproduce your case with no text and then not being able to type anything :-/ Check if it is still present after the last change.

alboiuvlad29:
@witwash Nope... It's still acting funny

https://user-images.githubusercontent.com/45460964/187140318-095079c7-0622-41d4-a1c9-effd4c675bec.mov

Checkout the branch attached to this PR(https://github.com/MewsSystems/tech-mo/pull/1418) and have a look.

witwash:
@alboiuvlad29 it was because removing the last character (and returning `TextEditingValue.empty`) would give up the focus of the input. Should be OK now.

optimus/lib/src/form/date_input_form_field.dart
```diff
@@ -0,0 +1,385 @@
...
+
+  /// Function to be called when the user submits the form.
+  final ValueChanged<DateTime?> onSubmitted;
+
+  /// The custom format for the output. See [OptimusCustomFormat] for more
+  /// details.
+  ///
+  /// If [customFormat] is not provided, the visual representation of the date
+  /// will be generated from the [format].
+  final OptimusCustomFormat? customFormat;
```
ookami-kb, line 45: Let's not allow custom formats. Again, we're building an opinionated library, not some highly-customizable general-purpose one.

ookami-kb (changes requested):

optimus/lib/src/form/date_formatter.dart
```diff
@@ -0,0 +1,280 @@
...
+    required this.allowedDigits,
+  }) {
+    _userInput = userInput;
+    _buildMaskMap();
+  }
+
+  final String mask;
+  final String placeholder;
+  final String? initValue;
+  final RegExp allowedDigits;
```
ookami-kb, line 30: Why should it be a parameter?

optimus/lib/src/form/date_formatter.dart
```diff
@@ -0,0 +1,280 @@
...
+///
+/// Mask could consist of the following characters:
+///
+/// - `#`: Digit
+///
+/// Example: `###-###-###`
+///
+/// The [_userInput] is used for tracking the user input and distinct it from
+/// the placeholder, which is used to fill empty places.
+class DateFormatter extends TextInputFormatter {
```
ookami-kb, line 14: Please add some tests for the class, the functionality looks pretty complex.

ookami-kb:
@witwash the PR is pretty big. Let's break it into multiple parts.

The first part can be the input formatter together with tests.

optimus/lib/src/form/date_formatter.dart
witwash: Since DateFormat is providing the [digitMatcher](https://api.flutter.dev/flutter/intl/DateFormat/digitMatcher.html), I thought it could be passed as a parameter for the formatter.

PR #253, opened 2022-09-01
feat(optimus): Add DateFormatter

witwash:
#### Summary

The first part of the RND-114990

Added `DateFormatter` with tests. 

#### Testing steps

- `DateFormatter` is covered with tests. 

#### Follow-up issues

- Will add the component itself in part 2.

#### Check during review

- Unused code removed.
- Build passing.

### DIFF
--- a/optimus/lib/src/form/date_formatter.dart
+++ b/optimus/lib/src/form/date_formatter.dart
@@ -84,9 +84,7 @@ class DateFormatter extends TextInputFormatter {
       final selectPosition = _getNextInputIndex(0);
       if (_isValidDigit(newText[0])) {
         resultText = _replaceCharAt(placeholder, selectPosition, newText[0]);
-        resultSelection = TextSelection.collapsed(
-          offset: selectPosition + 1,
-        );
+        resultSelection = TextSelection.collapsed(offset: selectPosition + 1);
       } else {
         return oldValue;
       }
@@ -94,9 +92,7 @@ class DateFormatter extends TextInputFormatter {
       if (newText.length > oldText.length) {
         if (_isComplete(oldText) ||
             !_isValidPosition(newSelectionStart) ||
-            !_isValidDigit(
-              newText[newSelectionStart - 1],
-            )) {
+            !_isValidDigit(newText[newSelectionStart - 1])) {
           return oldValue;
         }
 
@@ -113,9 +109,7 @@ class DateFormatter extends TextInputFormatter {
               offset: _getNextInputIndex(nextInputSpace + 1),
             );
           } else {
-            resultSelection = TextSelection.fromPosition(
-              TextPosition(offset: nextInputSpace),
-            );
+            resultSelection = TextSelection.collapsed(offset: nextInputSpace);
           }
         } else {
           resultText = _replaceCharAt(
@@ -148,9 +142,7 @@ class DateFormatter extends TextInputFormatter {
             placeholder.substring(start, end),
           );
 
-          resultSelection = TextSelection.collapsed(
-            offset: selectionPosition,
-          );
+          resultSelection = TextSelection.collapsed(offset: selectionPosition);
         } else if (_isDesignatedSpace(newSelectionStart)) {
           final prevInputSpace = _getPreviousInputIndex(newSelectionStart);
           if (_isValidDigit(oldText[prevInputSpace])) {
@@ -171,10 +163,7 @@ class DateFormatter extends TextInputFormatter {
       }
     }
 
-    return TextEditingValue(
-      text: resultText,
-      selection: resultSelection,
-    );
+    return TextEditingValue(text: resultText, selection: resultSelection);
   }
 }
 
--- a/optimus/lib/src/form/date_formatter.dart
+++ b/optimus/lib/src/form/date_formatter.dart
@@ -1,4 +1,3 @@
-import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 
 /// TextInputFormatter that will format the input according to the given mask.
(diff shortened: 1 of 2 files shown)


optimus/test/date_formatter_test.dart
```diff
@@ -0,0 +1,934 @@
...
+void main() {
+  late TextEditingValue actualOldValue;
+  late TextEditingValue actualNewValue;
+  void callBack(TextEditingValue oldValue, TextEditingValue newValue) {
+    actualOldValue = oldValue;
+    actualNewValue = newValue;
+  }
+
+  group('Adding first valid value', () {
+    testWidgets('Adding valid input', (tester) async {
```
ookami-kb, line 62: Tests can be simplified. You don't need widget tests at all here, you're testing just a formatting functionality. Something like this:

```dart
test('valid input', () {
  final sut = DateFormatter(
    mask: '##-##-####',
    placeholder: 'DD-MM-YYYY',
    userInput: [],
  );

  const oldValue = TextEditingValue(
    text: '',
    selection: TextSelection.collapsed(offset: 0),
  );
  const newValue = TextEditingValue(
    text: '1',
    selection: TextSelection.collapsed(offset: 1),
  );
  final actual = sut.formatEditUpdate(oldValue, newValue);

  expect(
    actual,
    const TextEditingValue(
      text: '1D-MM-YYYY',
      selection: TextSelection.collapsed(offset: 1),
    ),
  );
});
```

optimus/lib/src/form/date_formatter.dart
```diff
@@ -0,0 +1,309 @@
...
+  /// # - is a place designed for a digit. Any
+  /// other character will be displayed as is.
+  ///
+  /// Example: '##-##-####'
+  final String mask;
+
+  /// The placeholder to fill the empty spaces in the mask.
+  ///
+  /// Example: 'DD-MM-YYYY'
+  final String placeholder;
```
ookami-kb, line 31: As a user, I shouldn't bother with both mask and placeholder. I should just enter the mask as `DD-MM-YYYY` and the widget should do everything else internally.

optimus/lib/src/form/date_formatter.dart
```diff
@@ -0,0 +1,309 @@
...
+
+  /// Digits that are allowed to be entered by the user.
+  final RegExp? allowedDigits;
```
ookami-kb, line 34: I don't think this is needed.

optimus/lib/src/form/date_formatter.dart
```diff
@@ -0,0 +1,309 @@
...
+
+  /// Map of indexes and the corresponding mask characters.
+  final Map<int, String> _mask = {};
+
+  /// Callback, when the user input changes.
+  final VoidCallback? onUserInputChanged;
```
ookami-kb, line 40: There should be no callback like this. This is a formatted, basically, pure function. It should just format the value.

optimus/lib/src/form/date_formatter.dart
```diff
@@ -0,0 +1,309 @@
...
+
+/// TextInputFormatter that will format the input according to the given mask.
+///
+/// DateFormatter will force the input to follow the mask and will fill the
+/// empty spaces with the placeholder.
+class DateFormatter extends TextInputFormatter {
+  DateFormatter({
+    required this.mask,
+    required this.placeholder,
+    required List<int> userInput,
```
ookami-kb, line 12: User shouldn't provide this, it should be done internally if needed.

ookami-kb (changes requested):

optimus/lib/src/form/date_formatter.dart
witwash: It does. You can pass the `DateFormat` to the widget and it will format it and will use it for the `DateFormatter`.

optimus/lib/src/form/date_formatter.dart
witwash: This is the list of positions that were input by the user. All of it is for the `OptimusDateInputField/FormField`, not for the end user.

witwash:
I reworked the DateInput component. Since we don't support letter input (web doesn't), it can be simplified. Should be more bearable now.

optimus/test/date_formatter_test.dart
```diff
@@ -0,0 +1,460 @@
...
+        actual,
+        const TextEditingValue(
+          text: '01-01-2020',
+          selection: TextSelection.collapsed(offset: 10),
+        ),
+      );
+    });
+  });
+
+  testWidgets('Starting with initial value', (tester) async {
```
ookami-kb, line 348: Why do you need widget tests at all for `DateFormatter`? It can just be unit tests, as it doesn't make sense to test the input field integration with input formatted – that's what Flutter does.

witwash: Good point. Will delete it.

optimus/test/date_formatter_test.dart
```diff
@@ -0,0 +1,430 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:optimus/src/form/date_formatter.dart';
+
+void main() {
```
ookami-kb, line 5: Let's further optimize tests code. You can add a helper method:

```dart
@isTest
void testFormat({
  required String format,
  TextEditingValue oldValue = const TextEditingValue(
    text: '',
    selection: TextSelection.collapsed(offset: 0),
  ),
  required TextEditingValue newValue,
  required TextEditingValue expected,
}) {
  test('formats ${newValue.text} according to $format', () {
    final formatter = DateFormatter(placeholder: format);

    expect(formatter.formatEditUpdate(oldValue, newValue), expected);
  });
}
```

And then rewrite tests to smth like that:

```dart
testFormat(
  format: 'DD-MM-YYYY',
  newValue: const TextEditingValue(
    text: '1',
    selection: TextSelection.collapsed(offset: 1),
  ),
  expected: const TextEditingValue(
    text: '1D-MM-YYYY',
    selection: TextSelection.collapsed(offset: 1),
  ),
);
```

ookami-kb:
@witwash I've done some refactoring. Tests are passing, but please check that I didn't break any logic.

witwash:
@ookami-kb, thank you.

ookami-kb (approved):

PR #258, opened 2022-09-15
feat!: Upgrade dart_code_metrics to 4.18.3

ookami-kb:
#### Summary

Upgraded `dart_code_metrics` to version `4.18.3` and fixed lint errors.

#### Testing steps

No, analyzer update.

#### Follow-up issues

No.

#### Check during review

- Unused code removed.
- Build passing.

### DIFF
--- a/optimus/lib/src/number_picker/number_picker.dart
+++ b/optimus/lib/src/number_picker/number_picker.dart
@@ -38,7 +38,7 @@ class OptimusNumberPickerFormField extends FormField<int> {
           enabled: enabled,
           autovalidateMode: autovalidateMode,
           builder: (FormFieldState<int> field) {
-            void _onChanged(int? value) {
+            void handleChanged(int? value) {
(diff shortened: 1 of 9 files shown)


optimus/lib/src/number_picker/number_picker.dart
```diff
@@ -38,7 +38,7 @@ class OptimusNumberPickerFormField extends FormField<int> {
           enabled: enabled,
           autovalidateMode: autovalidateMode,
           builder: (FormFieldState<int> field) {
-            void _onChanged(int? value) {
+            void handleChanged(int? value) {
```
chaeMil, line 41: Why its suddenly public?

ookami-kb: It's not public, it's a local function and they shouldn't have `_` prefix.

chaeMil (approved):

PR #260, opened 2022-09-16
chore(release): publish packages

alboiuvlad29:
- mews_pedantic@0.9.0
 - optimus@0.19.0

#### Summary

#### Testing steps

#### Follow-up issues

#### Check during review

- Unused code removed.
- Build passing.

### DIFF
--- a/mews_pedantic/CHANGELOG.md
+++ b/mews_pedantic/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 0.9.0
+
+> Note: This release has breaking changes.
+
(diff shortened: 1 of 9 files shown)


mews_pedantic/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 0.9.0
```
ookami-kb, line 1: It shouldn't be there, `mews_pedantic` is already upgraded.

ookami-kb (changes requested):

PR #262, opened 2022-09-23
fix(optimus): [RND-122387] Fix onChange behaviour for the DateInput component.

witwash:
RND-122387

#### Summary

`onChange` wasn't working as expected. Now it would return `null` until the date is fully entered.

#### Testing steps

1. Override the `onChange` callback
2. Test if it would return `null` on an incomplete date (and after clearing) and will return the `DateTime` on a complete date.

#### Follow-up issues

Nope.

#### Check during review

- Unused code removed.
- Build passing.


alboiuvlad29 (approved):

kamil-grzadziela (approved):

witwash:
@ookami-kb, can you please look at this? 🙏

ookami-kb (approved):

PR #267, opened 2022-10-03
fix(optimus): [RND-120913] OptimusSelectFormField not opening when tapping on the chevron

witwash:
RND-120913

#### Summary

- fixed behavior when tapping on chevron would not open the `SelectInput` component.
- created the base class, because the `SelectInput` component was based on the `Search` component, which made little logical sense. Now there is a common class `DropdownSelect`, that is used by `SelectInput` and `Search` components.
- added `trailingImplicit` parameter. For example, this would be a chevron for the `SelectInput` and `Search` icon for the `Search` component. `trailingImplicit` widget is meant to control the dropdown, where the `trailing` action widget is defined by the user.
- added animation to the `SelectInput` chevron (same as in other dropdown components with chevron, mirroring the web version).
- reorganization of the file structure, because of the new common class.
#### Testing steps

1. Open Storybook. Open SelectInput story.
2. SelectInput will open the dropdown when clicking on the chevron icon.
3. Opening/closing the SelectInput will animate the chevron icon.

#### Follow-up issues

None

#### Check during review

- Unused code removed.
- Build passing.


chaeMil (approved):
LGTM

ookami-kb (approved):

PR #275, opened 2022-11-02
feat(kiosk_mode): [RND-125108] Implement `startKioskMode` on iOS

alboiuvlad29:
#### Summary

RND-125108

This PR adds implementations for `startKioskMode` and `stopKioskMode` on iOS.

- `stopKioskMode` signature changed to allow querying wether the response was successful or not from iOS and for Android is changed as well
- `watchKioskMode` will be more reactive. 

From Apple's documentation, the method will work when:
- Device is supervised in Apple Business Manager
- App using this package is allowed to run under Single App Mode

>You can use this method to lock your app into Single App mode and to release it from that mode later. For example, a test-taking app might enter this mode at the beginning of a test and exit it when the user completes the test. Entering Single App mode is supported only for devices that are supervised using Mobile Device Management (MDM), and the app itself must be enabled for this mode by MDM. You must balance each call to enter Single App mode with a call to exit that mode.
Because entering or exiting Single App mode might take some time, this method executes asynchronously and notifies you of the results using the completionHandler block.

https://developer.apple.com/documentation/uikit/uiaccessibility/1615186-requestguidedaccesssession

#### Testing steps

Unfortunately, without having a supervised device, there isn't really a way to test this.  
Functionality was dev-tested with the response from the `completionHandler` mocked.
```swift
 requestGuidedAccessSession(enabled: true, completionHandler: { result($0) })
```
```swift
 requestGuidedAccessSession(enabled: true, completionHandler: { result(!$0) })
```

#### Follow-up issues

No.

#### Check during review

- Unused code removed.
- Build passing.

### DIFF
--- a/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt
+++ b/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt
@@ -54,7 +54,6 @@ class KioskModePlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
                 }
             }
         } ?: result.success(false)
-
     }
 
     private fun stopKioskMode(result: MethodChannel.Result) {
--- a/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt
+++ b/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt
@@ -55,7 +55,7 @@ class KioskModePlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
 
     private fun stopKioskMode(result: MethodChannel.Result) {
         activity?.stopLockTask()
-        result.success(isInKioskMode() ?: false)
+        result.success(true)
     }
 
 
--- a/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt
+++ b/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt
@@ -54,7 +54,9 @@ class KioskModePlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
     }
 
     private fun stopKioskMode(result: MethodChannel.Result) {
-        val isInKioskMode = isInKioskMode() ?: false
+        // it may happen that isInKioskMode returns null when the service could not be found
+        // in this case, it's better to allow calling stopLockTask() than being stuck in kiosk mode
+        val isInKioskMode = isInKioskMode() ?: true
 
         if(isInKioskMode) {
             activity?.stopLockTask()
--- a/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt
+++ b/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt
@@ -33,6 +33,7 @@ class KioskModePlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
             "startKioskMode" -> startKioskMode(result)
             "stopKioskMode" -> stopKioskMode(result)
             "isInKioskMode" -> isInKioskMode(result)
+            "isManagedKiosk" -> isManagedKiosk(result)
             else -> result.notImplemented()
         }
     }
@@ -55,6 +56,17 @@ class KioskModePlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
     private fun stopKioskMode(result: MethodChannel.Result) {
         activity?.stopLockTask()
         result.success(true)
+    }
+
+
+    private fun isManagedKiosk(result: MethodChannel.Result) {
+        val service = activity?.getSystemService(Context.ACTIVITY_SERVICE) as? ActivityManager
+        if (service == null) {
+            result.success(null)
+            return
+        }
+
+        result.success(service.lockTaskModeState == ActivityManager.LOCK_TASK_MODE_LOCKED)
     }
 
     private fun isInKioskMode(result: MethodChannel.Result) {
--- a/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt
+++ b/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt
@@ -33,7 +33,6 @@ class KioskModePlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
             "startKioskMode" -> startKioskMode(result)
             "stopKioskMode" -> stopKioskMode(result)
             "isInKioskMode" -> isInKioskMode(result)
-            "isManagedKiosk" -> isManagedKiosk(result)
             else -> result.notImplemented()
         }
     }
@@ -55,18 +54,7 @@ class KioskModePlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
 
     private fun stopKioskMode(result: MethodChannel.Result) {
         activity?.stopLockTask()
-        result.success(null)
-    }
-
-
-    private fun isManagedKiosk(result: MethodChannel.Result) {
-        val service = activity?.getSystemService(Context.ACTIVITY_SERVICE) as? ActivityManager
-        if (service == null) {
-            result.success(null)
-            return
-        }
-
-        result.success(service.lockTaskModeState == ActivityManager.LOCK_TASK_MODE_LOCKED)
+        result.success(true)
     }
 
     private fun isInKioskMode(result: MethodChannel.Result) {
--- a/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt
+++ b/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt
@@ -54,15 +54,17 @@ class KioskModePlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
     }
 
     private fun stopKioskMode(result: MethodChannel.Result) {
-        // it may happen that isInKioskMode returns null when the service could not be found
-        // in this case, it's better to allow calling stopLockTask() than being stuck in kiosk mode
-        val isInKioskMode = isInKioskMode() ?: true
-
-        if(isInKioskMode) {
+        fun stopLockTask() {
             activity?.stopLockTask()
             result.success(true)
-        } else {
-            result.success(false)
+        }
+
+        when(isInKioskMode()) {
+            // it may happen that isInKioskMode returns null when the service could not be found
+            // in this case, it's better to allow calling stopLockTask() than being stuck in kiosk mode
+            null -> stopLockTask()
+            true -> stopLockTask()
+            false -> result.success(false)
         }
     }
 
--- a/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt
+++ b/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt
@@ -48,7 +48,7 @@ class KioskModePlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
                 try {
                     a.startLockTask()
                     result.success(true)
-                    kioskModeHandler.sendUpdate(true)
+                    kioskModeHandler.setKioskModeState(true)
                 } catch (e: IllegalArgumentException) {
                     result.success(false)
                 }
@@ -61,7 +61,7 @@ class KioskModePlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
         fun stopLockTask() {
             activity?.stopLockTask()
             result.success(true)
-            kioskModeHandler.sendUpdate(false)
+            kioskModeHandler.setKioskModeState(false)
         }
 
         when(isInKioskMode()) {
--- a/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt
+++ b/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt
@@ -58,6 +58,7 @@ class KioskModePlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
     }
 
     private fun stopKioskMode(result: MethodChannel.Result) {
+        activity?.stopLockTask()
         result.success(null)
         kioskModeHandler.emit()
     }
--- a/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt
+++ b/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt
@@ -55,7 +55,7 @@ class KioskModePlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
 
     private fun stopKioskMode(result: MethodChannel.Result) {
         activity?.stopLockTask()
-        result.success(true)
+        result.success(isInKioskMode() ?: false)
     }
 
 
--- a/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt
+++ b/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt
@@ -54,8 +54,14 @@ class KioskModePlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
     }
 
     private fun stopKioskMode(result: MethodChannel.Result) {
-        activity?.stopLockTask()
-        result.success(true)
+        val isInKioskMode = isInKioskMode() ?: false
+
+        if(isInKioskMode) {
+            activity?.stopLockTask()
+            result.success(true)
+        } else {
+            result.success(false)
+        }
     }
 
 
--- a/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt
+++ b/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt
@@ -19,13 +19,15 @@ class KioskModePlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
     private lateinit var channel: MethodChannel
     private lateinit var eventChannel: EventChannel
     private var activity: Activity? = null
+    private lateinit var kioskModeHandler: KioskModeStreamHandler
 
     override fun onAttachedToEngine(flutterPluginBinding: FlutterPlugin.FlutterPluginBinding) {
         channel = MethodChannel(flutterPluginBinding.binaryMessenger, methodChannelName)
         channel.setMethodCallHandler(this)
+        kioskModeHandler = KioskModeStreamHandler(this::isInKioskMode)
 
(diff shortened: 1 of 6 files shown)


Sevastyan:
@alboiuvlad29 could you please join forces with @BegovicAmar and find out whether our IT would be able to help testing it with a real Apple Business Manager? If I'm not mistaken, they already have it for company's purposes. And I think it would be wise to have at least one iPad with MDM in QA.

kamil-grzadziela (approved):

alboiuvlad29:
@Sevastyan This cannot be tested separately, in a package, we need to test it in an app. 
Yes we can set up some devices in MDM but for that, we also need a release build.

So I suggest: 
- Let's focus on merging the fictionality in this package that enables us to use it
- Let's create a feature flag and implement this in Kiosk
- We will create a release, and ask for help from IT team to: 1. Help us supervise an iPad from the QA team, 2. Add Kiosk app to allowlist to automatically toggle single app mode.

You can see exactly the same code used in Apple's video here https://developer.apple.com/videos/play/wwdc2022/10152/, seek to 08:40.

Sevastyan (approved):
LGTM.

alboiuvlad29:
ping @ookami-kb

ookami-kb:
> This cannot be tested separately, in a package, we need to test it in an ap

@alboiuvlad29 why not? There is an example app there.

ookami-kb:
> BREAKING stopKioskMode signature changed to allow querying wether the response was successful or not from iOS

Switching from `void` to `bool` is not actually a breaking change.

> BREAKING isManagedKiosk was removed as deemed unnecessary

Why? We don't use it, but it can be used by someone else. It doesn't require big efforts to maintain it, so I don't see a reason to remove it.

kiosk_mode/example/lib/main.dart
```diff
@@ -37,16 +37,44 @@ class _HomeState extends State<_Home> {
   Widget build(BuildContext context) => Column(
         mainAxisSize: MainAxisSize.min,
         children: [
-          if (Platform.isAndroid) ...[
-            const MaterialButton(
-              onPressed: startKioskMode,
-              child: Text('Start Kiosk Mode'),
+          MaterialButton(
+            onPressed: () => startKioskMode().then(
+              (value) => ScaffoldMessenger.of(context).showSnackBar(
```
Sevastyan, line 42: Let's move `then` closure out to some private method. It will ease the reading and comprehension of the UI.

kiosk_mode/example/lib/main.dart
```diff
@@ -37,16 +37,44 @@ class _HomeState extends State<_Home> {
...
+              ),
             ),
-            const MaterialButton(
-              onPressed: stopKioskMode,
-              child: Text('Stop Kiosk Mode'),
+            child: const Text('Start Kiosk Mode'),
+          ),
+          MaterialButton(
+            onPressed: () => stopKioskMode().then(
+              (value) => ScaffoldMessenger.of(context).showSnackBar(
```
Sevastyan, line 58: Same here, let's move it to a private method.

kiosk_mode/example/lib/main.dart
```diff
@@ -37,16 +37,44 @@ class _HomeState extends State<_Home> {
...
+                SnackBar(
+                  content: Text(
+                    value
+                        ? 'Kiosk mode stopped.'
+                        : 'Kiosk mode could not be stopped or wasn\'t active to begin with.',
```
Sevastyan, line 63: Can we disable the button then, if we are not in kiosk mode?

Sevastyan, line 63: Can we disable the button then, if we are not in kiosk mode?

Sevastyan, line 63: Can we disable the button then, if we are not in kiosk mode?

Sevastyan, line 63: Can we disable the button then, if we are not in kiosk mode?

Sevastyan (changes requested):

kiosk_mode/example/lib/main.dart
```diff
@@ -37,16 +37,44 @@ class _HomeState extends State<_Home> {
...
+          MaterialButton(
+            onPressed: () => startKioskMode().then(
+              (value) => ScaffoldMessenger.of(context).showSnackBar(
+                SnackBar(
+                  content: Text(
+                    value
+                        ? 'Kiosk mode started'
+                        : Platform.isIOS
+                            ? 'Single App mode is supported only for devices that are supervised using Mobile Device Management (MDM) and the app itself must be enabled for this mode by MDM.'
+                            : 'Kiosk mode could not be started.',
```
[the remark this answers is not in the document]
alboiuvlad29: Can MDM actually forbid setting lock task mode on Android? The only times this could be `false` on Android it's when:
1. Activity is null
2. IllegalArgumentException is thrown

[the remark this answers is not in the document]
Sevastyan: If that's the case, maybe we should add "try again" into that message. So, it's not treated as "it's impossible".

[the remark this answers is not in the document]
alboiuvlad29: I've added this to the message.

kiosk_mode/example/lib/main.dart
```diff
@@ -33,34 +33,63 @@ class _Home extends StatefulWidget {
 class _HomeState extends State<_Home> {
   late final Stream<KioskMode> _currentMode = watchKioskMode();
 
+  Future<void> _showSnackBar(String message) async {
+    ScaffoldMessenger.of(context)
+        .showSnackBar(SnackBar(content: Text(message)));
+  }
```
Sevastyan, line 39: Why `_showSnackBar` is async? Why not to utilize one line for 

```dart
ScaffoldMessenger.of(context).showSnackBar(SnackBar(content: Text(message)));
```
?

kiosk_mode/example/lib/main.dart
```diff
@@ -33,34 +33,63 @@ class _Home extends StatefulWidget {
...
+            children: [
+              MaterialButton(
+                onPressed: mode == KioskMode.enabled
+                    ? null
+                    : () => startKioskMode()
+                        .then(
+                          (didStart) => didStart
+                              ? 'Kiosk mode started'
+                              : Platform.isIOS
+                                  ? 'Single App mode is supported only for devices that are supervised using Mobile Device Management (MDM) and the app itself must be enabled for this mode by MDM.'
```
Sevastyan, line 58: Let's follow some human-friendly line limit. Long literal Strings can be split:

```dart
: Platform.isIOS
    ? 'Single App mode is supported only for devices that are supervised using'
        ' Mobile Device Management (MDM) and the app itself must be enabled for this mode'
        ' by MDM.'
    : 'Kiosk mode could not be started. Please try again.',
```

kiosk_mode/example/lib/main.dart
```diff
@@ -33,34 +33,63 @@ class _Home extends StatefulWidget {
...
+          return Column(
+            mainAxisSize: MainAxisSize.min,
+            children: [
+              MaterialButton(
+                onPressed: mode == KioskMode.enabled
+                    ? null
+                    : () => startKioskMode()
+                        .then(
+                          (didStart) => didStart
+                              ? 'Kiosk mode started'
```
Sevastyan, line 56: I don't like this. The information about the current mode is already shown by the last Widget (`Current mode: `).

kiosk_mode/example/lib/main.dart
```diff
@@ -33,34 +33,63 @@ class _Home extends StatefulWidget {
...
+                              : Platform.isIOS
+                                  ? 'Single App mode is supported only for devices that are supervised using Mobile Device Management (MDM) and the app itself must be enabled for this mode by MDM.'
+                                  : 'Kiosk mode could not be started. Please try again.',
```
Sevastyan, line 59: Let's not tell anything at all, if we cannot tell anything useful. Don't show the message on Android (actually, on any platform that is not iOS).

kiosk_mode/example/lib/main.dart
```diff
@@ -33,34 +33,63 @@ class _Home extends StatefulWidget {
...
+              MaterialButton(
+                onPressed: mode == KioskMode.disabled
+                    ? null
+                    : () => stopKioskMode()
+                        .then(
+                          (didStop) => didStop
+                              ? 'Kiosk mode stopped.'
+                              : 'Kiosk mode could not be stopped or wasn\'t active to begin with.',
+                        )
+                        .then(_showSnackBar),
```
Sevastyan, line 73: Revert changes for "Stop Kiosk Mode" button.

kiosk_mode/example/lib/main.dart
```diff
@@ -33,34 +33,63 @@ class _Home extends StatefulWidget {
...
+                child: const Text('Stop Kiosk Mode'),
+              ),
+              MaterialButton(
+                onPressed: () => isManagedKiosk()
+                    .then((isManaged) => 'Kiosk is managed: $isManaged')
+                    .then(_showSnackBar),
+                child: const Text('Is Managed Kiosk'),
```
Sevastyan, line 80: `Check if managed` to keep the language consistent with `Check mode`.

kiosk_mode/example/lib/main.dart
```diff
@@ -33,34 +33,63 @@ class _Home extends StatefulWidget {
...
+              ),
+              MaterialButton(
+                onPressed: () => getKioskMode()
+                    .then((mode) => 'Kiosk mode: $mode')
+                    .then(_showSnackBar),
+                child: const Text('Check mode'),
+              ),
+              Text(
+                'Current mode: ${snapshot.data}',
+              ),
```
Sevastyan, line 90:
```dart
Text('Current mode: ${snapshot.data}'),
```
please.

Sevastyan (changes requested):

kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModeStreamHandler.kt
```diff
@@ -33,4 +36,6 @@ class KioskModeStreamHandler(val getKioskModeState: () -> Boolean?) : EventChann
         timer?.cancel()
         timer = null
     }
+
+    fun setKioskModeState(isInKioskMode: Boolean?) =  postEvent(isInKioskMode)
```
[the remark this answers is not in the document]
alboiuvlad29: I'm against exposing this outside of this class. `KioskModeStreamHandler.kt` has only one job: to notify the platform channel of changes in  kiosk mode, let it either be automatic (from timer query) or manual when the switch happened. It's not good to expose this outside of the class where it could be misused. Someone could call `endOfStream` preliminary.

kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt
```diff
@@ -46,16 +48,29 @@ class KioskModePlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
...
-        result.success(null)
+        fun stopLockTask() {
+            activity?.stopLockTask()
+            result.success(true)
+            kioskModeHandler.setKioskModeState(false)
+        }
+
+        when(isInKioskMode()) {
+            // it may happen that isInKioskMode returns null when the service could not be found
+            // in this case, it's better to allow calling stopLockTask() than being stuck in kiosk mode
```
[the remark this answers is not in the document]
alboiuvlad29: It doesn't make sense, as this would be a regression for iOS - since we can know wether the request was successful or not. On plugin side, someone might want to know wether the request was successful or not. Then it's the duty of the native implementation to respond accordingly. Since Android doesn't provide a way to check this, but reading the documentation the call to stop kiosk mode would fail if it wasn't in kiosk mode to begin with, we can assume the logic above: If the app was in kiosk mode, then it must succeed. Otherwise it would fail

@ookami-kb thoughts?

kiosk_mode/example/lib/main.dart
alboiuvlad29: This is meant to show the usage of this method, it's just an example, after all. `Current mode` shows the behaviour of the stream. Some people perhaps would not rely on that, so showing it like this in the example app is not problematic in any way.

kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt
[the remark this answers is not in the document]
ookami-kb: From the user's point of view, I would expect that the result of `stopKioskMode` is `true` if I'm not in kiosk mode anymore after this call. From that point of view, returning `false` if I'm not in kiosk mode, seems weird, it looks as if the call failed and I'm still in kiosk mode.

And yes, if we don't know the result of the call, we can return `null` as a result.

But it's fine if iOS returns strict `bool` if we **can know** the result there.

kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModeStreamHandler.kt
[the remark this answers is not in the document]
Sevastyan: Well, it can be misused even currently - someone can just send whatever he wants down the stream - either `true`, or `false` and it could be even not matching the reality. Let's refactor to this:

<details>
<summary>diff</summary>

```diff
diff --git a/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt b/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt
index 0dbcedc..fc4dbde 100644
--- a/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt
+++ b/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt
@@ -48,7 +48,7 @@ class KioskModePlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
                 try {
                     a.startLockTask()
                     result.success(true)
-                    kioskModeHandler.setKioskModeState(true)
+                    kioskModeHandler.emit()
                 } catch (e: IllegalArgumentException) {
                     result.success(false)
                 }
@@ -61,7 +61,7 @@ class KioskModePlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
         fun stopLockTask() {
             activity?.stopLockTask()
             result.success(true)
-            kioskModeHandler.setKioskModeState(false)
+            kioskModeHandler.emit()
         }
 
         when(isInKioskMode()) {
diff --git a/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModeStreamHandler.kt b/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModeStreamHandler.kt
index 3291b2b..9c74b72 100644
--- a/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModeStreamHandler.kt
+++ b/kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModeStreamHandler.kt
@@ -7,7 +7,7 @@ import java.util.*
 import kotlin.concurrent.fixedRateTimer
 
 
-class KioskModeStreamHandler(val getKioskModeState: () -> Boolean?) : EventChannel.StreamHandler {
+class KioskModeStreamHandler(private val getKioskModeState: () -> Boolean?) : EventChannel.StreamHandler {
     private var eventSink: EventChannel.EventSink? = null
     private var timer: Timer? = null
     private var previousState: Boolean? = getKioskModeState()
@@ -17,16 +17,7 @@ class KioskModeStreamHandler(val getKioskModeState: () -> Boolean?) : EventChann
         val period: Int = (arguments as Map<*, *>)["androidQueryPeriod"] as Int
         eventSink = events
         timer = fixedRateTimer(period = period.toLong(), daemon = true) {
-           postEvent(getKioskModeState())
-        }
-    }
-
-    private fun postEvent(currentState: Boolean?) {
-        if (currentState != previousState) {
-            previousState = currentState
-            eventSink?.let {
-                mainHandler.post { it.success(currentState) }
-            }
+            emit()
         }
     }
 
@@ -37,5 +28,13 @@ class KioskModeStreamHandler(val getKioskModeState: () -> Boolean?) : EventChann
         timer = null
     }
 
-    fun setKioskModeState(isInKioskMode: Boolean?) =  postEvent(isInKioskMode)
+    fun emit() {
+        val currentState = getKioskModeState()
+        if (currentState != previousState) {
+            previousState = currentState
+            eventSink?.let {
+                mainHandler.post { it.success(currentState) }
+            }
+        }
+    }
 }
```

</details>

kiosk_mode/android/src/main/kotlin/com/mews/kiosk_mode/KioskModePlugin.kt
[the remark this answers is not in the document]
Sevastyan: @alboiuvlad29 Please, revert it.This file is for Android. And on Android it should always be `result.success(null)` - as it was previously.

And I agree with Kirill's comment - I don't expect `false`, if I wasn't in kiosk mode before this call. My intention is to stop kiosk mode, so when I call `stopKioskMode` I should get `true` for when the mode was actually stopped, or nothing happened because there was nothing to stop - in both cases I expect `true` to be returned.

kiosk_mode/lib/kiosk_mode.dart
```diff
@@ -18,23 +18,35 @@ enum KioskMode {
...
+///  Otherwise, the result will be `false`. Note: [Activity.stopLockTask][1] does not guaranteed LockTask
+/// or screen pinning mode will be stopped.
 ///
-/// It's only supported on Android currently, where it calls `stopLockTask`.
-/// It will throw `FlutterMethodNotImplemented` on iOS.
-Future<void> stopKioskMode() => _channel.invokeMethod('stopKioskMode');
+/// On iOS, the result will be `true` if the app was in Single App mode, `false` - otherwise.
+///
+/// [1]: https://developer.android.com/reference/android/app/Activity#stopLockTask()
+Future<bool> stopKioskMode() => _channel
```
Sevastyan, line 47: Make the return `bool` nullable and update the documentation comment. On Android, it should just fire `Activity.stopLockTask()` and return `.success(null)`.

kiosk_mode/example/lib/main.dart
Sevastyan: The quality of the code should not depend on whether this is just an example, or something intended for production. If anything, the quality of an example should be elevated, since it's intended for public use.

> Some people perhaps would not rely on that

What do you mean?

My problem with all this is that the information is duplicated.

Sevastyan (approved):
Changing of `stopKioskMode` signature - in this case changing the return type from `void` to `bool` is not a breaking change. Please, update the description and PR title.

Sevastyan:
@ookami-kb it's ready to be reviewed by you 🙏🏻

kiosk_mode/example/lib/main.dart
```diff
@@ -33,34 +33,63 @@ class _Home extends StatefulWidget {
...
+        ' using Mobile Device Management (MDM) and the app itself must'
+        ' be enabled for this mode by MDM.',
+      );
+    }
+  }
+
+  void _handleStop(bool? didStop) {
+    if (didStop == false) {
+      _showSnackBar(
+        'Kiosk mode could not be stopped or was not active to begin with.',
```
ookami-kb, line 52:
```suggestion
        'Kiosk mode could not be stopped or was not active.',
```

ookami-kb (approved):

PR #276, opened 2022-11-14
feat(optimus)!: [RND-126039] Additional changes for date input field

alboiuvlad29:
#### Summary

See usage in this PR: https://github.com/MewsSystems/tech-mo/pull/1576.

Fixes issue when input from android native keyboard wasn't working.
Allows picker to be changed by provided `value`.

Marking as breaking since `initialValue` has been renamed to `value`.

Thanks @witwash for your help!

#### Testing steps

No.

#### Follow-up issues

No.
#### Check during review

- Unused code removed.
- Build passing.

### DIFF
--- a/optimus/lib/src/form/date_input_field.dart
+++ b/optimus/lib/src/form/date_input_field.dart
@@ -73,18 +73,21 @@ class OptimusDateInputField extends StatefulWidget {
   State<OptimusDateInputField> createState() => _OptimusDateInputFieldState();
 }
 
-class _OptimusDateInputFieldState extends State<OptimusDateInputField> with ThemeGetter {
+class _OptimusDateInputFieldState extends State<OptimusDateInputField>
+    with ThemeGetter {
   StyledInputController? _styleController;
   FocusNode? _focusNode;
   String _previousValue = '';
 
-  StyledInputController get _controller => _styleController ??= StyledInputController(
+  StyledInputController get _controller =>
+      _styleController ??= StyledInputController(
         text: _formatValue(widget.value),
         inputStyle: _inputStyle,
         placeholderStyle: _placeholderStyle,
       );
 
-  FocusNode get _effectiveFocusNode => widget.focusNode ?? (_focusNode ??= FocusNode());
+  FocusNode get _effectiveFocusNode =>
+      widget.focusNode ?? (_focusNode ??= FocusNode());
 
   TextStyle get _placeholderStyle => theme.getPlaceholderStyle(widget.size);
 
@@ -101,7 +104,8 @@ class _OptimusDateInputFieldState extends State<OptimusDateInputField> with Them
     super.didUpdateWidget(oldWidget);
     if (oldWidget.value != widget.value) {
       _updateControllerValue(widget.value);
-    } else if (oldWidget.format.pattern != widget.format.pattern || oldWidget.format.locale != widget.format.locale) {
+    } else if (oldWidget.format.pattern != widget.format.pattern ||
+        oldWidget.format.locale != widget.format.locale) {
       final oldDate = _getDateTime(oldWidget.format, _controller.text);
       _updateControllerValue(oldDate);
     }
@@ -110,7 +114,9 @@ class _OptimusDateInputFieldState extends State<OptimusDateInputField> with Them
   void _onFocusChanged() {
     if (!_effectiveFocusNode.hasFocus) {
       widget.onEditCompleted?.call(
-        _controller.isInputComplete ? _getDateTime(widget.format, _controller.text) : null,
+        _controller.isInputComplete
+            ? _getDateTime(widget.format, _controller.text)
+            : null,
       );
     }
   }
@@ -136,11 +142,14 @@ class _OptimusDateInputFieldState extends State<OptimusDateInputField> with Them
     return result;
   }
 
-  String _formatValue(DateTime? value) => value != null ? _formatOutput(value) : '';
+  String _formatValue(DateTime? value) =>
+      value != null ? _formatOutput(value) : '';
 
   String _onChanged(String value) {
     if (_previousValue != value || value.isEmpty) {
-      final result = _controller.isInputComplete ? _getDateTime(widget.format, value) : null;
+      final result = _controller.isInputComplete
+          ? _getDateTime(widget.format, value)
+          : null;
       widget.onChanged?.call(result);
       _previousValue = value;
     }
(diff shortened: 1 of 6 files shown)


witwash:
I've updated the name since there is a breaking change.

witwash (approved):

alboiuvlad29:
ping @ookami-kb  🙏🏻

optimus/lib/src/form/input_field.dart
```diff
@@ -106,6 +109,9 @@ class OptimusInputField extends StatefulWidget {
   final bool readOnly;
 
   final VoidCallback? onTap;
+  final VoidCallback? onClear;
+
+  final ValueChanged<bool>? onKeyboardVisibilityChanged;
```
ookami-kb, line 114: Why do you need this parameter?

alboiuvlad29: It was initially used to update the field value when the keyboard disappeared, but I've updated it to use focus nodes instead.

optimus/lib/src/form/date_input_field.dart
```diff
@@ -65,35 +76,59 @@ class OptimusDateInputField extends StatefulWidget {
...
 
   StyledInputController get _controller =>
       _styleController ??= StyledInputController(
-        text: _initialText,
+        text: _formatValue(widget.value),
         inputStyle: _inputStyle,
         placeholderStyle: _placeholderStyle,
       );
 
+  FocusNode get _effectiveFocusNode =>
```
ookami-kb, line 89: It's not attached to any widget, i.e. it's not used.

optimus/lib/src/form/date_input_field.dart
```diff
@@ -36,15 +39,23 @@ class OptimusDateInputField extends StatefulWidget {
...
+  final DateTime? value;
 
   /// The widget size. By default [OptimusWidgetSize.large].
   final OptimusWidgetSize size;
 
   /// {@macro flutter.widgets.editableText.onChanged}
   final ValueChanged<DateTime?>? onChanged;
 
+  /// Function to be called when the input focus was lost.
+  final ValueChanged<DateTime?>? onEditCompleted;
```
ookami-kb, line 52: I don't think that it should be done like this.

First, this is confusing, as there's `onEditingComplete` parameter in default Flutter widgets that behaves differently. 

Second, relying on the focus node doesn't give you the expected behavior. E.g. pressing some button or checkbox will not unfocus the current field.

optimus/lib/src/form/date_input_field.dart
```diff
@@ -181,8 +212,8 @@ class _OptimusDateInputFieldState extends State<OptimusDateInputField>
 
   @override
   void dispose() {
-    _controller.dispose();
     _styleController?.dispose();
+    _effectiveFocusNode.dispose();
```
ookami-kb, line 216: No, it should never be done like this. You cannot dispose the focus node that was passed as the parameter – you don't own it. Disposing should only be called for internal `_focusNode`, for external nodes only listeners should be removed.

ookami-kb (changes requested):

optimus/lib/src/form/date_input_field.dart
alboiuvlad29: I know, and I would never have done it this way, although for some reason `OptimusInputField` behaves the same so I was just keeping the behaviour similar. I can remove this and we can later on update the widgets to dispose focus nodes by the owner, but this would require changes in all packages using any widget with `OptimusInputField` as a child to remove the expectation that the child will handle disposing parent-provided arguments that need to be disposed.

optimus/lib/src/form/date_input_field.dart
```diff
@@ -36,15 +38,20 @@ class OptimusDateInputField extends StatefulWidget {
...
+  /// The value of the input.
+  final DateTime? value;
 
   /// The widget size. By default [OptimusWidgetSize.large].
   final OptimusWidgetSize size;
 
   /// {@macro flutter.widgets.editableText.onChanged}
   final ValueChanged<DateTime?>? onChanged;
 
+  /// Function to be called when the input focus was lost or keyboard is hid.
```
Sevastyan, line 50: `hid.`? Let's not shorten words.

optimus/lib/src/form/date_input_field.dart
```diff
@@ -36,15 +38,20 @@ class OptimusDateInputField extends StatefulWidget {
...
+  final VoidCallback? onClear;
```
Sevastyan, line 51: I don't think the name makes sense, given the things this callback is supposed to be called for, as per the documentation comment.

witwash (approved):

optimus/lib/src/form/input_field.dart
```diff
@@ -217,6 +219,7 @@ class _OptimusInputFieldState extends State<OptimusInputField>
   void _onClearAllTap() {
     _effectiveController.clear();
     widget.onChanged?.call('');
+    widget.onClear?.call();
```
witwash, line 222: Is `onChanged('')` not sufficient?

Sevastyan (commented):
@witwash Should it validate the input? Like, check that it is actually a date that makes sense?

* `line 9`: it is not a `FormField`.

@alboiuvlad29 I'm failing to understand what this PR is about. Could you describe the change in the PR description, please?

optimus/lib/src/form/date_input_field.dart
alboiuvlad29: Has been removed.

optimus/lib/src/form/date_input_field.dart
alboiuvlad29: Has been removed.

optimus/lib/src/form/date_input_field.dart
witwash: it's not a shortened word, it's my bad grammar :-/

witwash:
@Sevastyan,

The date is validated on if it's a `DateTime` or not (otherwise it's `null`). It's not validated whether it's a valid date in some certain context. You can validate it with `onSubmit` or `onChange`. We can add a range for the validation if it's necessary.

alboiuvlad29:
@Sevastyan Updated the comment. The PR adds 2 things:
1. A fix for native android keyboard, where input wasn't recognised (workaround is commented in code)
2. Allows the field to display values provided by `value` field.

Sevastyan (approved):
LGTM

alboiuvlad29:
@ookami-kb 🙏🏻

ookami-kb (approved):

PR #299, opened 2023-01-23
feat(storybook): Updated storybook version. Added saving theme.

witwash:
#### Summary

Updated storybook dependency. Added theme saving.

#### Testing steps

1. Open Storybook.
2. Change the theme.
3. Hot-reload - theme should be saved.

You can test it on the firebase [here](https://witwash-storybook.web.app/#/).

#### Follow-up issues

None.

#### Check during review

- Unused code removed.
- Build passing.


storybook/lib/main.dart
```diff
@@ -51,70 +52,128 @@ import 'package:storybook_flutter/storybook_flutter.dart';
...
-          loaderStory,
-          dateTimeFieldStory,
-          notificationStory,
-        ],
+  State<MyApp> createState() => _MyAppState();
+}
+
+class _MyAppState extends State<MyApp> {
+  final Future<SharedPreferences> _prefs = SharedPreferences.getInstance();
+  late Future<ThemeMode> _themeMode;
```
ookami-kb, line 64: Can it be final?

storybook/lib/main.dart
```diff
@@ -51,70 +52,128 @@ import 'package:storybook_flutter/storybook_flutter.dart';
...
+
+  @override
+  void initState() {
+    super.initState();
+    _themeMode = _getThemeModeFromPreferences();
+  }
+
+  Future<ThemeMode> _getThemeModeFromPreferences() async {
+    final prefs = await _prefs;
+    final value = prefs.getString('themeMode');
```
ookami-kb, line 74: Extract `themMode` to constants.

storybook/lib/main.dart
```diff
@@ -51,70 +52,128 @@ import 'package:storybook_flutter/storybook_flutter.dart';
...
+
+    return value.toThemeMode;
+  }
+
+  Future<void> _saveThemeMode(ThemeMode themeMode) async {
+    final prefs = await _prefs;
+    await prefs.setString('themeMode', themeMode.toString());
```
ookami-kb, line 81: I would rather be saving just `name` – it's a little bit more flexible, since doesn't rely on `toString()` implementation and the name of the class.

ookami-kb (changes requested):

storybook/lib/main.dart
```diff
@@ -54,67 +55,108 @@ void main() => runApp(const MyApp());
...
+                  notificationStory,
+                ],
+              );
+          }
+        },
       );
 }
+
+extension on String? {
+  ThemeMode get toThemeMode {
```
ookami-kb, line 148: You can remove hard-coded strings as well with smth like

```dart
ThemeMode.values.firstWhere((e) => e.name == this, orElse: () => ThemeMode.system);
```

storybook/lib/main.dart
```diff
@@ -54,67 +55,98 @@ void main() => runApp(const MyApp());
...
+                  notificationStory,
+                ],
+              );
+          }
+        },
       );
 }
+
+extension on String? {
+  ThemeMode get toThemeMode => ThemeMode.values
```
ookami-kb, line 148: nitpicking: `toThemeMode` sounds like a method name, so it should rather be `toThemeMode()`.

storybook/pubspec.yaml
```diff
@@ -11,6 +11,7 @@ dependencies:
     sdk: flutter
   optimus:
     path: ../optimus
+  shared_preferences: ^2.0.16
```
ookami-kb, line 14: Sorry, just noticed another small issue: shouldn't `.lock` file be updated as well?

witwash: https://github.com/MewsSystems/mews-flutter/blob/d94cec7b25865f6669059fe85536d613895b1614/.gitignore#L4 ¯\_(ツ)_/¯

ookami-kb (approved):

PR #316, opened 2023-02-08
feat(optimus)!: [RND-159990] Updated icons library

witwash:
#### Summary

Removed `card_keyhole` icon. 

#### Testing steps

None. Icons library update.

#### Follow-up issues

None.

#### Check during review

- Unused code removed.
- Build passing.


witwash:
fluttericon.com had an expired SSL certificate and was not usable. It's up now, but we might need to come up with an alternative solution for the future. I could use Glyphs 2 Mini and edit font files manually or something like this 🤔 (it's paid tho). I can't use FontForge (a free alternative) on macOS, because the package is not signed and policies forbid me to run it.

ookami-kb (approved):

PR #325, opened 2023-02-27
fix(optimus): [RND-164340] `OptimusNestedCard` does not have contrasting colour in dark mode

alboiuvlad29:
#### Summary

RND-164340

Changes the dark mode color to be more suitable on an OptimusCard in dark mode.

| Before | After |
| ------| ------ |
| image | image |

#### Testing steps

According to the issue.

#### Follow-up issues

#### Check during review

- Unused code removed.
- Build passing.

### DIFF
--- a/optimus/lib/src/card.dart
+++ b/optimus/lib/src/card.dart
@@ -156,7 +156,7 @@ class OptimusNestedCard extends StatelessWidget {
     switch (variant) {
       case OptimusNestedCardVariant.emphasized:
         return theme.isDark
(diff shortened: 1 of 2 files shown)


witwash:
@alboiuvlad29,
I've added new color and for the sake of consistency changed the opacity to 24%. Should look the same. 

Also, since it's our own initiative and there is no documentation on which color we should use, I've noted this component for re-review (to check if what we used is fine). If DS team will come up with any changes, I'll let you know.

witwash (approved):

PR #332, opened 2023-03-22
feat(optimus): [RND-168345] Updated OptimusSearch

witwash:
#### Summary

- Added a clear button for the `OptimusSearch`
- Updated docs
- Updated story

#### Testing steps

- Open the Search Field story and change the input
- Clear button should be visible only when the input is not empty.

#### Follow-up issues

None.

#### Check during review

- Unused code removed.
- Build passing.


witwash:
@ookami-kb,
can you please look at this one? I'd like to publish a new version, so `commander` could use an updated search component.

ookami-kb (approved):

PR #336, opened 2023-03-27
fix(optimus): [RND-169480] Fix clearAll button not reacting in the OptimusSearch

witwash:
[RND-169480]
#### Summary

- Dropdown overlay was conflicting with the clear button's GestureDetector.
- The overlay was entering the zone first (it's on the top of the screen) and was winning over the gesture 100% of the time. `CustomRawGestureDetector` will ensure the `clearAll` button will call `onTap`.

<details><summary>Video</summary>

https://user-images.githubusercontent.com/9210422/229759937-3e7092a4-5237-4d79-8b3c-88e90d418a18.mp4

</details> 

#### Testing steps

1. Open Search Input story
2. Enable `clearAll`
3. Try clearing the value, when the field is in focus.

#### Follow-up issues

- I've found that there is a [reference](https://github.com/MewsSystems/tech-mo/blob/0d406e58bea3e336f80476a67e2a0b6909693b1e/commander/integration_test/actions/actions_common.dart#L44) to the old key value in the tests. It should be updated to the `OptimusDropdownOverlay`. It's something that I've forgot to change after the consolidation. 
#### Check during review

- Unused code removed.
- Build passing.


optimus/lib/src/dropdown/dropdown_select.dart
```diff
@@ -76,10 +77,14 @@ class _DropdownSelectState<T> extends State<DropdownSelect<T>> {
   final _fieldBoxKey = GlobalKey();
 
   FocusNode? _focusNode;
+  TextEditingController? _controller;
```
ookami-kb, line 80: It's not disposed of.

ookami-kb (changes requested):

ookami-kb (approved):

PR #340, opened 2023-04-04
refactor(optimus): Update OptimusSearch story

witwash:
#### Summary

- `isUpdating` and `showLoader` were doing the same. 
-  I've removed the `showLoader`, but forgot to mark it as a breaking change.
- This PR has an updated story and is marked as breaking to reflect the change in the `OptimusSearch`.

#### Testing steps

None. Refactor.

#### Follow-up issues

None.

#### Check during review

- Unused code removed.
- Build passing.


ookami-kb:
@witwash as discussed, let's not mark it as breaking change, since it's already happened and the package was published.

ookami-kb (approved):

PR #346, opened 2023-04-23
feat(optimus): [RND-164749] Add OptimusCompactStepBar

witwash:
[RND-164749] 

#### Summary

- Added an `OptimusCompactStepBar`.
- Some code refactoring because of the reused components (`step_bar.dart` and `step_bar_item.dart` are without changes, it is a result of the split of one file into two).
- Added a Story for the `OptimusCompactStepBar`.

https://user-images.githubusercontent.com/9210422/233872672-2b88880b-e093-4119-b19e-66e2c59c84f8.mp4

</details>

#### Testing steps

1. Open the `Compact Step Bar` story.
2. Try it with different configurations.

#### Follow-up issues

This has to be on top of everything when expanded. Might need to add a fallback option if the list is too big. Thinking about using a dialog or something with scrollable content. Open to suggestions.

#### Check during review

- Unused code removed.
- Build passing.


optimus/lib/src/step_bar/step_bar.dart
```diff
@@ -0,0 +1,151 @@
...
+  active,
+
+  /// The step is inactive and unfinished.
+  enabled,
+
+  /// The step is disabled and unavailable.
+  disabled,
+}
+
+class OptimusStepBarItem {
```
ookami-kb, line 138: I think, it's better to extract these helper classes to a separate file, since they are shared between common and compact bars.

ookami-kb (changes requested):

witwash:
@ookami-kb, ping

PR #347, opened 2023-04-24
fix(optimus): [RND-170692] Dialog is overlapping with screen safe area

witwash:
[RND-170692]

#### Summary

- Added a possibility to wrap a dialog into the `SafeArea` widget. By default, it is not wrapped (as it was before). You can change it where it is needed.

CleanShot 2023-04-24 at 04 40 24

</details>

#### Testing steps

None

#### Follow-up issues

None

#### Check during review

- Unused code removed.
- Build passing.


ookami-kb:
Is there a case when it shouldn't be wrapped into `SafeArea`?

witwash:
@ookami-kb, no, don't think so, it has to respect the SafeArea. I wanted it to remain flexible, in case if forced SafeArea would break the design or something.

ookami-kb (approved):

PR #357, opened 2023-05-03
fix(optimus): [RND-174939] Fix focus change on SelectInput

witwash:
#### Summary

- Focus change was making select showing/hiding repeatedly because of the keyboard show. I've fixed it.
- Fixed embedded search losing focus on keyboard show/rebuild.

#### Testing steps

- Open `Storybook` on the mobile device
- Test the `SelectInput` component story with and without embedded search enabled.

#### Follow-up issues

None

#### Check during review

- Unused code removed.
- Build passing.


optimus/lib/src/dropdown/dropdown.dart
```diff
@@ -222,9 +229,9 @@ class _SearchWrapper extends StatelessWidget {
     return SizedBox(
       child: Column(
         children: [
-          if (showDivider && isOnTop) _buildDivider(theme),
-          child,
-          if (showDivider && !isOnTop) _buildDivider(theme),
+          if (widget.showDivider && widget.isOnTop) _buildDivider(theme),
+          Container(key: _searchKey, child: widget.child),
```
ookami-kb, line 233: I guess, `Container` is only used to add the key? You can use `KeyedSubtree` instead in that case.

ookami-kb (approved):

PR #358, opened 2023-05-09
refactor: Integrate DCM Teams

ookami-kb:
#### Summary

#### Testing steps

#### Follow-up issues

#### Check during review

- Unused code removed.
- Build passing.


mews_pedantic/lib/analysis_options.yaml
```diff
@@ -1,69 +1,103 @@
...
+    - avoid_types_as_parameter_names
+    - avoid_web_libraries_in_flutter
+    - cancel_subscriptions
+    - close_sinks
+    - collection_methods_unrelated_type
+    - comment_references
+    - control_flow_in_finally
+    - empty_statements
+    - hash_and_equals
+    # - implicit_reopen
```
witwash, line 21: Do we need commented rules here?

optimus/lib/src/dialogs/dialog_wrapper.dart
```diff
@@ -49,6 +51,12 @@ class _DialogWrapperState extends State<DialogWrapper>
     }
   }
 
+  @override
```
witwash, line 54: Good spot. Thank you

optimus/lib/src/tokens/tokens_color_dark.dart
```diff
@@ -8,7 +8,7 @@
 import 'dart:ui';
 
 class DesignTokensColor {
-  DesignTokensColor._();
+  const DesignTokensColor._();
```
witwash, line 11: I'll pass it to the DS team. We'll need to update the generator ;)

witwash:
LGTM.
I'll wait for others to share their opinions on new DCM rules in case somebody has their 2 cents to add.

mews_pedantic/lib/analysis_options.yaml
ookami-kb: They are rules for Dart 3. Right now, they give warnings since the current version of analyzer doesn't support them, but as soon as we migrate to Dart 3, I want to enable them.

alboiuvlad29 (approved):
LGTM. 👍🏻

witwash (approved):

witwash:
@Sevastyan, sorry, didn't mean to close before your review. Hope all those changes are OK with you as well

PR #359, opened 2023-05-12
feat(optimus): Allow for customization of `OptimusDateInputField`

Sevastyan:
#### Summary

This PR allows one to build his own date picker but with the same mask and its behaviour as defined by `OptimusDateInputField`. We want to use it in the new design of Mews Kiosk app. Below is how it looks:

grab

#### Testing steps

Test the date picker.

#### Follow-up issues

None.

#### Check during review

- Unused code removed.
- Build passing.


Sevastyan:
Not needed anymore.

PR #371, opened 2023-06-12
refactor!: Migration to Flutter 3.10

witwash:
#### Summary

- removed invalid rule

#### Testing steps

#### Follow-up issues

#### Check during review

- Unused code removed.
- Build passing.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -19,7 +19,7 @@ jobs:
 
       - uses: subosito/flutter-action@v1
         with:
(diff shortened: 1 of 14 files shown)


witwash:
@ookami-kb, ping

witwash:
@ookami-kb, can we please merge this one? It is blocking the 3.10 migration 🙏

ookami-kb:
Let's make it Dart 3 only in that case + enable the commented rules.

PR #372, opened 2023-06-14
feat: [RND-177241] Enable theming using tokens

witwash:
#### Summary

- Updated `OptimusTheme` and added an `OptimusTokens tokens` property, which will hold all tokens for light/dark theme.
- `OptimusTokens` is basically a `ThemeExtension` and I'm using `tailor` for the code generation.
- `OptimusTokens` could be overwritten with `copyWith`, this will enable custom theming. You need to wrap your stuff with `OptimusTheme` and provide your custom `OptimusTokens` for light and dark themes.

#### Testing steps

None. A prerequired PR for later tokens integration directly into components.

#### Follow-up issues
- I had to override `analyzer` to `5.12.0`, because they have introduced a new type - `InvalidType` and broke some parts of the code generation. `dynamic` types would resolve to `InvalidType` and the current (latest) `theme_tailor` package is not supporting it yet. `freezed` has some problems with it as well.
- Testing?
- More code generation, from the design-tokens PR, to make this process less error-prone.

#### Check during review

- Unused code removed.
- Build passing.

### DIFF
--- a/optimus/lib/src/theme/optimus_tokens.dart
+++ b/optimus/lib/src/theme/optimus_tokens.dart
@@ -242,6 +242,11 @@ class _$OptimusTokens {
     DesignTokensColorDark.borderInteractiveSecondaryActive
   ];
 
+  static List<Color> borderInteractiveSecondaryDefault = [
+    DesignTokensColorLight.borderInteractiveSecondaryDefault,
+    DesignTokensColorDark.borderInteractiveSecondaryDefault
+  ];
+
   static List<Color> borderInteractiveSecondaryHover = [
     DesignTokensColorLight.borderInteractiveSecondaryHover,
     DesignTokensColorDark.borderInteractiveSecondaryHover
--- a/optimus/lib/src/theme/optimus_tokens.dart
+++ b/optimus/lib/src/theme/optimus_tokens.dart
@@ -14,39 +14,1258 @@ class _$OptimusTokens {
     DesignTokensColorLight.backgroundAccentPrimary,
     DesignTokensColorDark.backgroundAccentPrimary
   ];
-
   static List<Color> backgroundAccentSecondary = [
     DesignTokensColorLight.backgroundAccentSecondary,
     DesignTokensColorDark.backgroundAccentSecondary
   ];
-
   static List<Color> backgroundAlertBasicPrimary = [
     DesignTokensColorLight.backgroundAlertBasicPrimary,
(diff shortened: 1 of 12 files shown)


witwash:
@ookami-kb I'd like to hear your thoughts before I continue working on it 🙏

witwash:
@ookami-kb, ping

ookami-kb:
Sorry, forgot to add a comment. Yeah, in general, it makes sense to me,

witwash:
`analyzer` dep is locked to `5.12.0`, cuz there is a breaking change in the minor update, and code generation is not working as intended.

optimus/pubspec.yaml
```diff
@@ -33,3 +35,6 @@ flutter:
     - family: OptimusIcons
       fonts:
         - asset: packages/optimus/fonts/OptimusIcons.ttf
+
+dependency_overrides:
+  analyzer: 5.12.0
```
ookami-kb, line 40: Is that a conflict because of `theme_tailor`? What do we need to remove this override?

optimus/lib/src/theme/optimus_tokens.dart
ookami-kb, line 1: Where do these data come from? Is it created manually?

optimus/lib/src/theme/optimus_tokens.tailor.dart
ookami-kb, line 1: Add this file to `.gitattributes` please.

ookami-kb (changes requested):

optimus/lib/src/theme/optimus_tokens.dart
witwash: I'm working on a script that will generate this one with the PR from the `design-tokens`. For now, yes, it was created manually.

optimus/pubspec.yaml
witwash: yep, it's because `theme_tailor` didn't update to fix the `analyzer` breaking change

witwash: I saw that `freezed` last update fixed that problem. I will check it out and create a PR to `theme_tailor` with the fix.

witwash: https://github.com/rrousselGit/freezed/pull/921, look like it was merged as a temporary fix before the author will come up with a better solution. I'll do my best, at worst I'll start a discussion about this bug ;)

ookami-kb (approved):

PR #376, opened 2023-06-19
refactor!: Migrate to Flutter 3.10

witwash:
#### Summary

- Migration to `Flutter 3.10`

#### Testing steps

None. Refactor.

#### Follow-up issues

None.

#### Check during review

- Unused code removed.
- Build passing.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -19,7 +19,7 @@ jobs:
 
       - uses: subosito/flutter-action@v1
         with:
-          flutter-version: "3.7.0"
+          flutter-version: "3.10.0"
 
       - name: Set up environment paths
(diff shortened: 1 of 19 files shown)


witwash:
@ookami-kb, can you please look at it?

witwash:
Also, I see that we have some analysis issues, that lower our score on pub.dev. I'll fix them here as well.

witwash:
@ookami-kb, ping

ookami-kb:
1. I don't think we should update `mews_pedantic` in the same PR as the other libraries. It's the breaking change in the package, let's release it separately.
2. I don't think we should force other libraries to support Dart 3+. It will prevent us from doing fixes until we migrate the apps code to Dart 3+, which will take some time.

PR #381, opened 2023-06-22
fix: [RND-178412] OptimusAvatar offline mode throwing exception

witwash:
RND-178412

#### Summary

- `OptimusAvatar` was throwing an exception when the load was unsuccessful. Now it won't (the error is handled, but now there is no unnecessary noise in the form of an exception in the log).
- I've replaced `CircleAvatar` with `FadeInImage` to add more animation when the image is loaded. Looks better. Before that, it would just pop in all at once.
- I've borrowed a `transparentImage` from the [transparent_image](https://pub.dev/packages/transparent_image), which he borrowed from Dart sources itself(it was private), so I guess we're even.

#### Testing steps

1. Open `Avatar` Story
2. Check is all avatars are displayed correctly.
3. Intentionally set an URL to fail and check if `OptimusAvatar` is set correctly.

#### Follow-up issues

None.

#### Check during review

- Unused code removed.
- Build passing.


witwash:
@ookami-kb ping

optimus/lib/src/avatar.dart
```diff
@@ -71,5 +112,75 @@ class _Indicator extends StatelessWidget {
   }
 }
 
+final Uint8List _transparentImage = Uint8List.fromList(<int>[
+  0x89,
```
ookami-kb, line 116: Make it const, and write in a more compact form with "manual" formatting, e.g.:

```dart
  0x89, 0x50, 0x4E, 0x47, 0x0D, 0x0A, 0x1A, //
  0x0A, 0x00, 0x00, 0x00, 0x0D, 0x49, 0x48,
  0x44, 0x52,
```

ookami-kb (changes requested):

ookami-kb (approved):

PR #424, opened 2023-08-28
build: Update Storybook dependencies

witwash:
#### Summary

- update Storybook deps

#### Testing steps

None

#### Follow-up issues

None

#### Check during review

- Verify against Jira issue.
- Unused code removed.
- Build passing.

### DIFF
--- a/storybook/pubspec.lock
+++ b/storybook/pubspec.lock
@@ -37,10 +37,10 @@ packages:
     dependency: transitive
     description:
       name: collection
-      sha256: "4a07be6cb69c84d677a6c3096fcf960cc3285a8330b4603e0d463d15d9bd934c"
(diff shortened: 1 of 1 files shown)


storybook/pubspec.lock
```diff
@@ -476,5 +476,5 @@ packages:
     source: hosted
     version: "1.0.0"
 sdks:
-  dart: ">=3.0.0 <4.0.0"
+  dart: ">=3.1.0-185.0.dev <4.0.0"
```
ookami-kb, line 479: Why should it depend on Dart `3.1.0`? If we allow Flutter `3.10`, Dart version `3.0.0` should be allowed as well.

ookami-kb (approved):

PR #432, opened 2023-09-11
feat!: Update rules set

witwash:
#### Summary

- update to DCM 1.9.0
- add new rules

### add new rules:
 - [avoid-missing-interpolation](https://dcm.dev/docs/rules/common/avoid-missing-interpolation/)
 - [avoid-passing-self-as-argument](https://dcm.dev/docs/rules/common/avoid-passing-self-as-argument/)
 - [avoid-recursive-calls](https://dcm.dev/docs/rules/common/avoid-recursive-calls/)
 - [avoid-unconditional-break](https://dcm.dev/docs/rules/common/avoid-unconditional-break/)
 - [avoid-unnecessary-if](https://dcm.dev/docs/rules/common/avoid-unnecessary-if/)
 - [move-variable-closer-to-its-usage](https://dcm.dev/docs/rules/common/move-variable-closer-to-its-usage/)
### rules that were added, but not enabled:
- [avoid-identical-exception-handling-blocks](https://dcm.dev/docs/rules/common/avoid-identical-exception-handling-blocks/) - not sure about this one, but we can enable it. Need another opinion.
- [avoid-passing-default-values](https://dcm.dev/docs/rules/common/avoid-passing-default-values/) - here as well, sometimes we pass the default value to ensure it would not break or something, but I think we should enable it and be more strict about breaking changes/changing the default value
- [avoid-weak-cryptographic-algorithms](https://dcm.dev/docs/rules/common/avoid-weak-cryptographic-algorithms/) - feels like out of the scope of this project
- [parameters-ordering](https://dcm.dev/docs/rules/common/parameters-ordering/)
- [prefer-getter-over-method](https://dcm.dev/docs/rules/common/prefer-getter-over-method/) - I feel like it would go against our naming scheme with methods/getters. Better to leave disabled

#### Testing steps

None

#### Follow-up issues

None

#### Check during review

- Verify against Jira issue.
- Unused code removed.
- Build passing.

### DIFF
--- a/optimus/lib/src/dropdown/dropdown.dart
+++ b/optimus/lib/src/dropdown/dropdown.dart
@@ -80,15 +80,6 @@ class _DropdownContent<T> extends StatelessWidget {
     );
   }
 
-  Widget? _buildEmptyPlaceholder() {
-    if (emptyResultPlaceholder != null) {
-      return Material(
-        color: Colors.transparent,
-        child: emptyResultPlaceholder,
-      );
-    }
-  }
-
   Widget _buildSearch(
     AnchoredOverlayController controller,
     bool isOnTop,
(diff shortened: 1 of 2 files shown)


witwash:
@ookami-kb, any idea why an action would target `1.8.4`, when I clearly defined `1.9.0` as a target? using `latest` (default value) is targeting 1.8.4 as well. Something with cache?

witwash:
@ookami-kb ping

ookami-kb:
> [avoid-identical-exception-handling-blocks](https://dcm.dev/docs/rules/common/avoid-identical-exception-handling-blocks/) - not sure about this one, but we can enable it. Need another opinion.

Yeah, seems a little bit artificial, let's leave it disabled for now.

> [avoid-passing-default-values](https://dcm.dev/docs/rules/common/avoid-passing-default-values/) - here as well, sometimes we pass the default value to ensure it would not break or something, but I think we should enable it and be more strict about breaking changes/changing the default value

I would say, these are different use cases: "I don't care about this parameter" (e.g. logging / default timeout) and "I need this exact value, and I don't care if it's a default one". Since it's up for the caller of this method to decide, I would leave this rule disabled.

> [avoid-weak-cryptographic-algorithms](https://dcm.dev/docs/rules/common/avoid-weak-cryptographic-algorithms/) - feels like out of the scope of this project

Yeah, but in general it makes sense to have this rule enabled, it won't hurt. I would enable it.

> [parameters-ordering](https://dcm.dev/docs/rules/common/parameters-ordering/)

Yeah, doesn't make sense in our code style.

> [prefer-getter-over-method](https://dcm.dev/docs/rules/common/prefer-getter-over-method/) - I feel like it would go against our naming scheme with methods/getters. Better to leave disabled

Agree.

witwash:
@ookami-kb, enabled `avoid-weak-cryptographic-algorithms`, anything else?

ookami-kb (changes requested):
Also, you will need to rebase the branch to keep breaking change commit for lint rules, and non-breaking for changes.

ookami-kb (approved):

PR #452, opened 2023-10-09
feat!: Update to DCM 1.10.0

witwash:
#### Summary

- Updated rules set
- Fixed linter issues 

##### Enabled rules:

- [avoid-local-functions](https://dcm.dev/docs/rules/common/avoid-local-functions/)
- [avoid-nullable-parameters-with-default-values](https://dcm.dev/docs/rules/common/avoid-nullable-parameters-with-default-values/)
- [avoid-extensions-on-records](https://dcm.dev/docs/rules/common/avoid-extensions-on-records/)
- [avoid-accessing-collections-by-constant-index](https://dcm.dev/docs/rules/common/avoid-accessing-collections-by-constant-index/)
- [avoid-unnecessary-getter](https://dcm.dev/docs/rules/common/avoid-unnecessary-getter/)
- [match-getter-setter-field-names](https://dcm.dev/docs/rules/common/match-getter-setter-field-names/)
- [avoid-wildcard-cases-with-enums](https://dcm.dev/docs/rules/common/avoid-wildcard-cases-with-enums/)
- [avoid-unnecessary-late](https://dcm.dev/docs/rules/common/avoid-unnecessary-late/)
- [avoid-single-child-column-or-row](https://dcm.dev/docs/rules/flutter/avoid-single-child-column-or-row/)
- [prefer-sliver-prefix](https://dcm.dev/docs/rules/flutter/prefer-sliver-prefix/)

##### Rules left disabled:

- [prefer-named-imports](https://dcm.dev/docs/rules/common/prefer-named-imports/)
- [avoid-long-files](https://dcm.dev/docs/rules/common/avoid-long-files/)

#### Testing steps

None

#### Follow-up issues

None

#### Check during review

- Verify against Jira issue.
- Unused code removed.
- Build passing.

### DIFF
--- a/.github/workflows/dcm.yml
+++ b/.github/workflows/dcm.yml
@@ -16,7 +16,7 @@ jobs:
         uses: CQLabs/setup-dcm@v1
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
-          version: '1.9.1'
+          version: '1.10.2'
 
       - uses: ./.github/actions/setup
 
--- a/mews_pedantic/lib/analysis_options.yaml
+++ b/mews_pedantic/lib/analysis_options.yaml
@@ -242,6 +242,7 @@ dart_code_metrics:
     # Common
 
     # - arguments-ordering
+    - avoid-accessing-collections-by-constant-index
     # - avoid-accessing-other-classes-private-members
     # - avoid-async-call-in-sync-function
     # - avoid-banned-file-names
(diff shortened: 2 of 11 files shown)


ookami-kb:
- `avoid-extensions-on-records` – I wouldn't turn it on, seems pretty useful. Basically, if extension on the record feels wrong, it means that even without extension record doesn't belong here.
- `avoid-local-functions` – I don't think that it complicates readability that much. The alternative is to have it in a bigger scope (class-private or file-private) which can be even more confusing if it's only used inside of one function.

PR #466, opened 2023-10-19
ci: Update Storybook deploy action trigger

witwash:
#### Summary

Trigger paths were not defined correctly and GH `Storybook Deploy` action would not start.

#### Testing steps

None

#### Follow-up issues

None

#### Check during review

- Verify against Jira issue.
- Unused code removed.
- Build passing.


witwash:
Deploy is not triggered on changes and I'd like to fix it. I believe this is the reason

ookami-kb (approved):

PR #470, opened 2023-11-09
feat!: Upgrade to DCM 1.11.0

witwash:
#### Summary

Update to DCM 1.11.1

##### Enabled rules:
- [avoid-duplicate-cascades](https://dcm.dev/docs/rules/common/avoid-duplicate-cascades/)
- [avoid-duplicate-switch-case-conditions](https://dcm.dev/docs/rules/common/avoid-duplicate-switch-case-conditions/)
- [avoid-unnecessary-super](https://dcm.dev/docs/rules/common/avoid-unnecessary-super/)
- [match-lib-folder-structure](https://dcm.dev/docs/rules/common/match-lib-folder-structure/)
- [prefer-both-inlining-annotations](https://dcm.dev/docs/rules/common/prefer-both-inlining-annotations/)
- [prefer-correct-handler-name](https://dcm.dev/docs/rules/common/prefer-correct-handler-name/)
- [prefer-correct-setter-parameter-name](https://dcm.dev/docs/rules/common/prefer-correct-setter-parameter-name/)
- [prefer-explicit-function-type](https://dcm.dev/docs/rules/common/prefer-explicit-function-type/)
- [prefer-explicit-type-arguments](https://dcm.dev/docs/rules/common/prefer-explicit-type-arguments/)
- [prefer-null-aware-spread](https://dcm.dev/docs/rules/common/prefer-null-aware-spread/)
- [prefer-specific-cases-first](https://dcm.dev/docs/rules/common/prefer-specific-cases-first/)
- [prefer-typedefs-for-callbacks](https://dcm.dev/docs/rules/common/prefer-typedefs-for-callbacks/)
- [prefer-unique-test-names](https://dcm.dev/docs/rules/common/prefer-unique-test-names/)

##### New rules that were left disabled:
- [avoid-banned-annotations](https://dcm.dev/docs/rules/common/avoid-banned-annotations/)
- [avoid-collection-mutating-methods](https://dcm.dev/docs/rules/common/avoid-collection-mutating-methods/)
- [avoid-inferrable-type-arguments](https://dcm.dev/docs/rules/common/avoid-inferrable-type-arguments/)
- [avoid-missing-test-files](https://dcm.dev/docs/rules/common/avoid-missing-test-files/)
- [prefer-addition-subtraction-assignments](https://dcm.dev/docs/rules/common/prefer-addition-subtraction-assignments/)
- [prefer-correct-callback-field-name](https://dcm.dev/docs/rules/common/prefer-correct-callback-field-name/)
- [prefer-prefixed-global-constants](https://dcm.dev/docs/rules/common/prefer-prefixed-global-constants/)

#### Testing steps

None

#### Follow-up issues

None

#### Check during review

- Verify against Jira issue.
- Unused code removed.
- Build passing.


witwash:
@ookami-kb, WDYT about enabling [prefer-correct-handler-name](https://dcm.dev/docs/rules/common/prefer-correct-handler-name/) and [prefer-correct-callback-field-name](https://dcm.dev/docs/rules/common/prefer-correct-callback-field-name)? I know that is something that we agreed on, but are we ready to enforce it? :D

witwash:
And what about [avoid-collection-mutating-methods](https://dcm.dev/docs/rules/common/avoid-collection-mutating-methods)? I think we could enable it with local only 🤔

witwash:
Well, I think we should enable `handler` one, but leave the `callback-field-name` turned off. We have plenty of `typedef` functions that are reported by this and I don't think that renaming the to `"on...." ` is making any sense.

ookami-kb:
> Well, I think we should enable `handler` one, but leave the `callback-field-name` turned off. We have plenty of `typedef` functions that are reported by this and I don't think that renaming the to `"on...." ` is making any sense.

Makes sense to me. @alboiuvlad29 @chaeMil @Sevastyan your input?

mews_pedantic/lib/analysis_options.yaml
```diff
@@ -386,7 +399,9 @@ dart_code_metrics:
         ignore-abstract: false
     - prefer-early-return
     # - prefer-enums-by-name
+    - prefer-explicit-function-type
     - prefer-explicit-parameter-names
+    # - prefer-explicit-type-arguments
```
ookami-kb, line 404: Let's enable this one instead of `avoid-inferrable-type-arguments`. I was specifically requesting for this rule for cases with `get_it` and `provider` where omitting type arguments can lead to confusing and fragile code.

mews_pedantic/lib/analysis_options.yaml
```diff
@@ -350,6 +357,7 @@ dart_code_metrics:
     # - map-keys-ordering
     # - match-class-name-pattern
     - match-getter-setter-field-names
+    # - match-lib-folder-structure
```
ookami-kb, line 360: I would enable this (yeah, it will have a lot of triggers in our code).

mews_pedantic/lib/analysis_options.yaml
```diff
@@ -368,15 +376,20 @@ dart_code_metrics:
     # - no-magic-number
     - no-object-declaration
     # - parameters-ordering
+    - prefer-addition-subtraction-assignments
```
ookami-kb, line 379: I don't think this rule is useful, I would disable it.

mews_pedantic/lib/analysis_options.yaml
```diff
@@ -397,15 +412,20 @@ dart_code_metrics:
...
       ignore-single: true
     # - prefer-named-imports
+    - prefer-null-aware-spread
     - prefer-parentheses-with-if-null
+    # - prefer-prefixed-global-constants
     - prefer-public-exception-classes
     - prefer-return-await
     - prefer-returning-conditional-expressions
     - prefer-simpler-patterns-null-check
+    # - prefer-specific-cases-first
```
ookami-kb, line 422: Looks useful, why disabling it?

ookami-kb (changes requested):

mews_pedantic/lib/analysis_options.yaml
witwash: Was stuck on the question "what is specific?" and had a feeling we would have been something like enum ordering enforce. I'll enable it and we'll see.

Sevastyan:
> > Well, I think we should enable `handler` one, but leave the `callback-field-name` turned off. We have plenty of `typedef` functions that are reported by this and I don't think that renaming the to `"on...." ` is making any sense.
> 
> Makes sense to me. @alboiuvlad29 @chaeMil @Sevastyan your input?

I agree. In most cases having `on....` for typedefs doesn't make sense. Let's ignore.

optimus/lib/src/typography/title.dart
```diff
@@ -1,3 +1,5 @@
+// ignore_for_file: prefer-correct-handler-name
```
ookami-kb, line 1: Why does it need to be ignored here?

witwash: It was triggering on the preset resolver. I've added a comment on why it is ignored there.

CleanShot 2023-11-14 at 12 24 13

ookami-kb: Discussed with DCM author, global functions will be excluded from this check in the next releases.

For now, let's ignore them inline instead of file-level ignoring, otherwise, this ignore won't be marked as obsolete when the rule is updated.

witwash: Done

ookami-kb (approved):

PR #472, opened 2023-11-21
ci: Bump Flutter to 3.16.0

witwash:
#### Summary

- bumped GH action Flutter version
- fixed migration issues

#### Testing steps

- `WillPopScope` is deprecated and replaced with the `PopScope`. There is an [issue](https://github.com/flutter/flutter/issues/138614) raised asking to preserve `WillPopScope`, but the outcome is not clear. Need to test if the back button press is acting as intended.

#### Follow-up issues

- Material 3 is now enabled by default and we need to test if it is not bleeding through our component and everything is looking like it should. There is a possibility to disable Material 3 for the time, but Material 2 is going to be deprecated so it would just postpone possible issues.
- #474 

#### Check during review

- Verify against Jira issue.
- Unused code removed.
- Build passing.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -6,7 +6,7 @@ runs:
     - name: Set up Flutter
       uses: subosito/flutter-action@ea686d7c56499339ad176e9f19c516ff6cf05a31
       with:
-        flutter-version: 3.13.9
+        flutter-version: 3.16.0
         cache: true
 
     - name: Set up environment paths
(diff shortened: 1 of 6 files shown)


ookami-kb:
There are also some lint rules updated in Dart 3.2, let's include them as well.

witwash:
@ookami-kb rules PR is [here](https://github.com/MewsSystems/mews-flutter/pull/474)

witwash:
@ookami-kb, ping

witwash:
@ookami-kb, didn't notice, thanks for noting. Done 👍

ookami-kb (approved):

PR #475, opened 2023-11-28
ci: Add tokens PR action

witwash:
#### Summary

Added an action that would trigger on push in `design-tokens-**` branch. This action will create a PR to `master`.

This action will ensure we have the latest version of tokens from [design-tokens](https://github.com/MewsSystems/design-tokens). Right now we have an action that is committing new files to `design-tokens-${version}` branch. This action will create a PR and generate files that were touched by the tokens update. 

#### Testing steps

We'll need to check whether the trigger is working correctly and the PR is created.

#### Follow-up issues

~~Set up an action that would regenerate `theme_tailor` files in this same PR.~~ done

#### Check during review

- Verify against Jira issue.
- Unused code removed.
- Build passing.

### DIFF
--- a/.github/workflows/tokens-pr.yml
+++ b/.github/workflows/tokens-pr.yml
@@ -1,0 +1,32 @@
+name: Create Design Tokens PR
+on:
+  push:
+    branches:
+      - "design-tokens-**"
+
+jobs:
+  pull-request:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the repository
+        uses: actions/checkout@v2
+
+      - name: Set up Flutter
+        uses: "./.github/actions/setup"
+
+      - name: Generate tokens files
+        run: melos gen_theme
+
+      - name: Commit changes
+        uses: stefanzweifel/git-auto-commit-action@v5
+        with:
+          commit_message: Automated commit of generated tailor files
+          file_pattern: "*tailor.dart"
+
+      - name: Create Pull Request
+        uses: repo-sync/pull-request@v2
+        with:
+          destination_branch: "master"
+          pr_title: "feat: Design Tokens Update"
+          pr_body: "Design Tokens were updated! This PR was created to update the code."
+          github_token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


witwash:
@ookami-kb, WDYT? 🤔

.github/workflows/tokens-pr.yml
```diff
@@ -0,0 +1,32 @@
...
+        run: melos gen_theme
+
+      - name: Commit changes
+        uses: stefanzweifel/git-auto-commit-action@v5
+        with:
+          commit_message: Automated commit of generated tailor files
+          file_pattern: "*tailor.dart"
+
+      - name: Create Pull Request
+        uses: repo-sync/pull-request@v2
```
ookami-kb, line 27: Let's rather use GitHub CLI, it's pre-installed on action runners: https://cli.github.com/manual/gh_pr_create

.github/workflows/tokens-pr.yml
```diff
@@ -0,0 +1,32 @@
...
+        uses: actions/checkout@v2
+
+      - name: Set up Flutter
+        uses: "./.github/actions/setup"
+
+      - name: Generate tokens files
+        run: melos gen_theme
+
+      - name: Commit changes
+        uses: stefanzweifel/git-auto-commit-action@v5
```
ookami-kb, line 21: Why do we need a separate action for that? Isn't it just a couple lines of code?

ookami-kb (changes requested):

.github/workflows/tokens-pr.yml
witwash: For no particular reason tbh. I wrote it with commands at first, then I looked at the action and it felt cleaner

ookami-kb: I would rather avoid using 3rd party actions unless they bring a lot of benefits.

.github/workflows/tokens-pr.yml
```diff
@@ -0,0 +1,36 @@
...
+
+      - name: Set up Flutter
+        uses: "./.github/actions/setup"
+
+      - name: Generate tokens files
+        run: melos gen_theme
+
+      - name: Commit changes
+        run: |
+          git config --local user.email "tech.account@mewssystems.com"
```
witwash, line 24: Do we have a user that we should use here? In `design-tokens` repo we use `tech.account@mewssystems.com`

PR #485, opened 2023-12-06
ci: Update tokens action

witwash:
#### Summary

- removed reviewers (assigned automatically based on the files changed)
- commit step would be skipped if `tailor.dart` file is not changed

#### Testing steps

Check if the action is working as intended.

#### Follow-up issues

None

#### Check during review

- Verify against Jira issue.
- Unused code removed.
- Build passing.

### DIFF
--- a/.github/workflows/tokens-pr.yml
+++ b/.github/workflows/tokens-pr.yml
@@ -21,12 +21,18 @@ jobs:
       - name: Generate tokens files
         run: melos gen_theme
 
+      - name: Check if tailor file was changed
+        id: tailor_changed
+        run: git status --porcelain | grep -q 'tailor.dart'
+        continue-on-error: true
+
       - name: Commit changes
+        if: steps.tailor_changed.outcome == 'success'
         run: |
           git config --local user.email "tech.account@mewssystems.com"
           git config --local user.name "mews-tech"
 
-          git add .
+          git add optimus/lib/src/theme/optimus_tokens.tailor.dart
           git commit -m "Update generated files"
 
           git push origin ${{ github.ref_name }}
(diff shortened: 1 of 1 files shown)


.github/workflows/tokens-pr.yml
```diff
@@ -21,12 +21,18 @@ jobs:
       - name: Generate tokens files
         run: melos gen_theme
 
+      - name: Check if tailor file was changed
+        id: tailor_changed
+        run: git status --porcelain | grep -q 'tailor.dart'
+        continue-on-error: true
+
       - name: Commit changes
+        if: steps.tailor_changed.outcome == 'success'
```
witwash, line 30: I think this is the right way, buy maybe you have a better idea 🤔

ookami-kb: I guess you can just use `success()` for checking all the previous steps without specifying id: https://docs.github.com/en/actions/learn-github-actions/expressions#status-check-functions

But shouldn't it also be applied to creating PR step?

witwash: No, this is for the case when the value itself has changed (in `tokens_dark/light.dart`), but the structure didn't (no removes/renames/additions). In that case, the `optimus_tokens.tailor.dart` file would be the same, but we still want the PR with new values.

ookami-kb: Ah, then yeah, makes sense.

ookami-kb (approved):

.github/workflows/tokens-pr.yml
ookami-kb: But cannot you just do `git add optimus/lib/src/theme/optimus_tokens.tailor.dart` always? If it's the same, it just won't be added, no?

witwash:
@ookami-kb,
so, can I merge this one? 🤔

ookami-kb (approved):

PR #488, opened 2023-12-06
ci: Update .lock files in dependabot PRs

witwash:
#### Summary

Added a follow-up action to commit `.lock` files if PR was created by the `dependabot`.

#### Testing steps

`Dependabot` PR should trigger this action and changes in `.lock` should be committed.

#### Follow-up issues

None

#### Check during review

- Verify against Jira issue.
- Unused code removed.
- Build passing.

### DIFF
--- a/.github/workflows/dependabot-follow-up.yml
+++ b/.github/workflows/dependabot-follow-up.yml
@@ -1,0 +1,35 @@
+name: Update .lock files on dependabot PR
+on:
+  pull_request:
+    types:
+      - opened
+jobs:
+  run_if:
+    if: startsWith(github.head_ref, 'dependabot/')
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.ref }}
+
+      - name: Set up Flutter
+        uses: ./.github/actions/setup
+
+      - name: Get dependencies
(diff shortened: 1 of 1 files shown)


witwash:
@ookami-kb,
looks like dependabot is updated and this one is not needed anymore

witwash:
@ookami-kb, what do you think?

.github/workflows/dependabot-follow-up.yml
```diff
@@ -0,0 +1,35 @@
+name: Update .lock files on dependabot PR
+on:
+  pull_request:
+    types:
+      - opened
```
ookami-kb, line 5: PR can also be updated/rebased, I think, it should run in this case as well.

ookami-kb (commented):
In general, yeah, makes sense, let's try it this way.

witwash:
If this [one](https://frontside.com/blog/2020-05-26-github-actions-pull_request/) is true,
> When a Github Actions workflow is configured to run on pull requests, by default it will be triggered on three different types of activities: `opened`, `synchronize`, and `reopened`.

 then by default, this action would be triggered on `open`, `reopen`, and `synchronize`. But I could not find it in the docs and decided to define it explicitly.

ookami-kb (approved):

PR #491, opened 2023-12-07
ci: Set to ignore local packages

witwash:
#### Summary

Added `mews_pedantic` to ignore in `storybook`'s `dependabot`. This will ensure `dependabot` will not switch `mews_pedantic` to `pub` version in its PRs.

#### Testing steps

Re-run dependabot and see if `mews_pedantic` dependency was changed.

#### Follow-up issues

None

#### Check during review

- Verify against Jira issue.
- Unused code removed.
- Build passing.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -24,7 +24,8 @@ updates:
   - package-ecosystem: "pub"
     directory: "/storybook/"
     ignore:
-      # mews_pedantic is a relative dependency that is always up to date
+      # relative dependencies that is always up to date
       - dependency-name: "mews_pedantic"
+      - dependency-name: "optimus"
     schedule:
       interval: "monthly"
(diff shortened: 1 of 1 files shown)


witwash:
@ookami-kb,
what about this one? Otherwise, dependabot will update it to the pub version. See #489 
CleanShot 2023-12-12 at 17 55 26

witwash:
@ookami-kb, ping

ookami-kb:
But in `pubspec.lock` these dependencies are still "locked" to the specific version, no? Shouldn't they be updated?

witwash:
@ookami-kb,
I'm a bit confused here. Yes, they are locked, but **dependabot** is changing it to the latest **hosted** one.
What I'm trying to do here is stop **dependabot** from changing `pubspec.lock` dependency that is pointing at the **local** package to the **hosted** one.

ookami-kb:
Let's discuss it in a call.

witwash:
Closing as it is indeed a hosted dependency, thus this PR is not needed.

PR #492, opened 2023-12-11
feat!: Upgrade to DCM 1.12.0

witwash:
#### Summary

Upgraded to DCM 1.12.0

#### Enabled rules: 
- [prefer-switch-with-enums](https://dcm.dev/docs/rules/common/prefer-switch-with-enums/)
- [avoid-misused-test-matchers](https://dcm.dev/docs/rules/common/avoid-misused-test-matchers/)
- [prefer-any-or-every](https://dcm.dev/docs/rules/common/prefer-any-or-every/)
- [avoid-duplicate-test-assertions](https://dcm.dev/docs/rules/common/avoid-duplicate-test-assertions/)

#### Disabled rules:
- [prefer-correct-handler-name](https://dcm.dev/docs/rules/common/prefer-correct-handler-name/) - too many false positives

#### Testing steps

None

#### Follow-up issues

None

#### Check during review

- Verify against Jira issue.
- Unused code removed.
- Build passing.

### DIFF
--- a/mews_pedantic/lib/analysis_options.yaml
+++ b/mews_pedantic/lib/analysis_options.yaml
@@ -266,6 +266,7 @@ dart_code_metrics:
     - avoid-duplicate-named-imports
     - avoid-duplicate-patterns
     - avoid-duplicate-switch-case-conditions
+    - avoid-duplicate-test-assertions
     # - avoid-dynamic
     - avoid-equal-expressions
     - avoid-explicit-pattern-field-name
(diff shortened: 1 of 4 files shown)


mews_pedantic/lib/analysis_options.yaml
ookami-kb, line 1: Let's also enable `ignore-parameters` config option for `prefer-typedefs-for-callbacks`.

ookami-kb (approved):

PR #494, opened 2023-12-13
ci: Fix checkout ref

witwash:
#### Summary

Fixed the ref name for follow-up action.

#### Testing steps

Rebase dependabot PR and check if the action would end successfully.

#### Follow-up issues

None

#### Check during review

- Verify against Jira issue.
- Unused code removed.
- Build passing.


witwash:
Should fix the problem that occurred in #489

ookami-kb (approved):

PR #504, opened 2024-01-04
feat: Add support for breaking releases

witwash:
#### Summary

- PRs created from the design tokens branch ending with `0.0` would be marked as `breaking`.

#### Testing steps

- New design tokens version should create a proper PR and depending on its version it should be marked as breaking.

#### Follow-up issues

None

#### Check during review

- Verify against Jira issue.
- Unused code removed.
- Build passing.

### DIFF
--- a/.github/workflows/tokens-pr.yml
+++ b/.github/workflows/tokens-pr.yml
@@ -6,6 +6,9 @@ on:
     paths:
       - "optimus/lib/src/theme/optimus_tokens.dart"
 
+env:
+  PR_TYPE: "feat"
+
 jobs:
   pull-request:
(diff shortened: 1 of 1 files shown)


ookami-kb:
Can it happen that after let's say `1.2.3` there will be `2.0.1`?

witwash:
I hope not 😀 well, I feel like if this would cover 90% of releases it's good. In case of a jump to a breaking version without 0.0 we would have to change the title manually. I was thinking about it and didn't want to overengineere. 

I'm open to suggestions 👐

ookami-kb (approved):

PR #531, opened 2024-01-29
feat: [DX-723] Migrate to typography tokens

witwash:
#### Summary

- removed old presets and replaced them with typography tokens
- updated components to use matching tokens (some of them are missing the definition in the design and I'll fixed them 1 by 1 in followups)
- updated stories
#### Testing steps

This PR is touching almost all components. Check the overall look of the storybook stories and I'll fix them 1 by 1 in the follow-ups if needed.

#### Follow-up issues

Some of the components were using styles not defined in design and I need to contact design team and update them accordingly.

#### Check during review

- Verify against Jira issue.
- Unused code removed.
- Build passing.


witwash:
Shoot, I merged it before seeing your comment :-/

witwash:
@ookami-kb, 
`presets.dart` wasn't exposed so it isn't breaking there. One thing that could be breaking is creating `optimusMaterialTheme` with its `baseTextStyle`, because `bodyMedium` has defined `fontSize` 🤔

PR #539, opened 2024-02-06
feat: [DX-1034] Add characters count to the OptimusInputField

witwash:
#### Summary

1. Added a character count track to the `OptimusInputField`
2. Updated story

<details><summary>Screenshots</summary></details>

#### Testing steps

1. Open Input story
2. Set the limit for the input characters
4. Test whether components corresponds with the design

#### Follow-up issues

None

#### Check during review

- Verify against Jira issue.
- Unused code removed.
- Build passing.


optimus/lib/src/common/field_wrapper.dart
```diff
@@ -187,14 +192,34 @@ class _FieldWrapper extends State<FieldWrapper> with ThemeGetter {
...
+                        overflow: TextOverflow.fade,
+                        style: TextStyle(
+                          color: captionColor,
+                        ),
+                        child: helperMessage,
+                      ),
+                    ),
+                  ),
+                if (helperMessage == null) const Spacer(),
+                if (controller != null && maxCharacters != null)
```
ookami-kb, line 217: That looks weird that in order to have char limit working, you need to pass `controller` as well.

witwash: I can just pass the current number of chars in the field instead. Or get it from the context as an alternative 🤔

ookami-kb: If it's easier with controller, it can use controller, but this class can create one if needed.

witwash: The `inputField` is passed to the wrapper as a `Widget`, so there is no ability to get an input length from there in another way than passing the controller or some of its values directly. Or wrap it with some inherited data with the controller and get it from the context.

optimus/lib/src/common/field_wrapper.dart
```diff
@@ -24,10 +25,12 @@ class FieldWrapper extends StatefulWidget {
     this.fieldBoxKey,
     this.children = const <Widget>[],
     this.size = OptimusWidgetSize.large,
+    this.maxCharacters,
```
ookami-kb, line 28: On further thought, should this wrapper know at all about `maxCharacters`? It should probably just accept `Widget? helpSuffix`, and it would be up to the input field to pass the character counter.

witwash: yeah, could be done this way as well. I'll change it 👍

witwash:
@ookami-kb, done 🤝

ookami-kb (approved):

PR #567, opened 2024-03-01
feat!: Upgrade to Flutter 3.19.2

witwash:
#### Summary

- **[BREAKING CHANGE]** minimal version is now `3.19.0`
- Changed CI Flutter version to `3.19.2`
- Fixed unused imports that were reported after the migration to `3.19.0`

#### Testing steps

Dev-tested.

#### Follow-up issues

Update the tokens template and remove `dart:ui` import.

#### Check during review

- Verify against Jira issue.
- Unused code removed.
- Build passing.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -6,7 +6,7 @@ runs:
     - name: Set up Flutter
       uses: subosito/flutter-action@ea686d7c56499339ad176e9f19c516ff6cf05a31
       with:
-        flutter-version: 3.16.9
+        flutter-version: 3.19.2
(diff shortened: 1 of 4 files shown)


witwash:
I guess the imports have changed and we no longer need to import `dart:ui` for the `FontFeature` class.

ookami-kb:
> I guess the imports have changed and we no longer need to import `dart:ui` for the `FontFeature` class.

Then it's rather breaking change for a library as well, we should set minimal Flutter version in `pubspec.yaml`.

ookami-kb (approved):

PR #572, opened 2024-03-08
fix: Fix MultiselectSelectInput layout

witwash:
#### Summary

- enabled SelectInput to expand in height if there are multiple selected values that would not fit.

#### Testing steps

1. Open Storybook Select Input story
2. Enable Multiselect
3. Select/deselect values and check if everything is working as intended.

#### Follow-up issues

None

#### Check during review

- Verify against Jira issue.
- Unused code removed.
- Build passing.


chaeMil:
@ookami-kb any updates on this? would be nice to merge this update to commander before releasing the Housekeeping assignment screen update.

ookami-kb (approved):

PR #595, opened 2024-04-29
feat: [DX-1521] Update icons library

witwash:
#### Summary

- added new icons:
  - `caret_filled_up`
  - `caret_filled_down`
- updated `All Icons` story

#### Testing steps

1. Open `All Icons` story
2. Check if the icons weren't messed up (for some reason sometimes `fluttericon.com` would generate the wrong `name: icon` pairs, so it needs to be double-checked).

#### Follow-up issues

None

#### Check during review

- Verify against Jira issue.
- Unused code removed.
- Build passing.


chaeMil (approved):
LGTM

PR #608, opened 2024-05-13
feat!: Upgrade to DCM 1.17.0

witwash:
#### Summary

Upgrade to DCM 1.17.0

##### Enabled rules:
- [avoid-collection-equality-checks](https://dcm.dev/docs/rules/common/avoid-collection-equality-checks/)
- [avoid-multi-assignment](https://dcm.dev/docs/rules/common/avoid-multi-assignment/)
- [prefer-switch-with-sealed-classes](https://dcm.dev/docs/rules/common/prefer-switch-with-sealed-classes/)

##### Rules left disabled:
- [prefer-single-declaration-per-file](https://dcm.dev/docs/rules/common/prefer-single-declaration-per-file/)

#### Testing steps

None, linter update.

#### Follow-up issues

None.

#### Check during review

- Verify against Jira issue.
- Unused code removed.
- Build passing.

### DIFF
--- a/kiosk_mode/example/lib/main.dart
+++ b/kiosk_mode/example/lib/main.dart
@@ -65,15 +65,18 @@ class _HomeState extends State<_Home> {
             mainAxisSize: MainAxisSize.min,
             children: [
               MaterialButton(
-                onPressed: mode == null || mode == KioskMode.enabled
-                    ? null
-                    : () => startKioskMode().then(_handleStart),
+                onPressed: switch (mode) {
+                  null || KioskMode.enabled => null,
+                  KioskMode.disabled => () =>
+                      startKioskMode().then(_handleStart)
+                },
                 child: const Text('Start Kiosk Mode'),
               ),
               MaterialButton(
-                onPressed: mode == null || mode == KioskMode.disabled
-                    ? null
-                    : () => stopKioskMode().then(_handleStop),
+                onPressed: switch (mode) {
+                  null || KioskMode.disabled => null,
+                  KioskMode.enabled => () => stopKioskMode().then(_handleStop),
+                },
                 child: const Text('Stop Kiosk Mode'),
               ),
               MaterialButton(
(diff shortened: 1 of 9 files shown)


ookami-kb:
@witwash links to the rules in PR description are broken.

alboiuvlad29:
`avoid-collection-equality-checks` we mostly use immutable lists everywhere which does deep equality comparison. Does this check consider that? If not we should probably disable it as it will be a pain to deal with this check afterwards.

alboiuvlad29:
I also don't see any benefit to `avoid-multi-assignment` as well. There are some places we do this on purpose.

ookami-kb:
@alboiuvlad29 I cannot find any usage of multi-assignment.

witwash:
I can see how neglecting `avoid-multi-assignment` can lead to less readable code 🤔. Who are you saving that one/two lines for, @alboiuvlad29? 😄

alboiuvlad29:
Never-mind, it was a different usage I had in mind and this isn't it regarding `avoid-multi-assignment`. Sorry about that. 😅

kiosk_mode/example/lib/main.dart
```diff
@@ -52,6 +52,16 @@ class _HomeState extends State<_Home> {
     }
   }
 
+  VoidCallback? _onStartPressed(KioskMode? mode) => switch (mode) {
```
ookami-kb, line 55: Inline these methods, otherwise, the name is confusing. It's not a handler method, it's a method that creates a handler.

ookami-kb (changes requested):

ookami-kb (approved):

witwash:
@chaeMil 👀

chaeMil (approved):

witwash:
@alboiuvlad29, 🙏

alboiuvlad29 (approved):

PR #628, opened 2024-06-18
ci: [DX-1806] Automate package publish on tag

witwash:
#### Summary

- added a workflow to publish `optimus` on tag pushed.

We need to set it up on the other [end](https://dart.dev/tools/pub/automated-publishing#configuring-automated-publishing-from-github-actions-on-pub-dev) to allow this automation.

#### Testing steps

- Create a release and try if it works, after all is set up.

#### Follow-up issues

Tune this workflow and do the same for other packages that need to be published on `pub.dev`

#### Check during review

- Verify against Jira issue.
- Unused code removed.
- Build passing.

### DIFF
--- a/.github/workflows/optimus-publish.yml
+++ b/.github/workflows/optimus-publish.yml
@@ -12,3 +12,4 @@ jobs:
     uses: dart-lang/setup-dart/.github/workflows/publish.yml@v1
     with:
       working-directory: optimus/
+      environment: pub.dev
--- a/.github/workflows/optimus-publish.yml
+++ b/.github/workflows/optimus-publish.yml
@@ -1,0 +1,14 @@
+name: Publish optimus to pub.dev
+
+on:
+  push:
+    tags:
+      - 'optimus-v[0-9]+.[0-9]+.[0-9]+*'
+
+jobs:
+  publish:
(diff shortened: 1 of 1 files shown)


witwash:
@ookami-kb, could you please enable it on the pub.dev admin side? 🙏 
https://dart.dev/tools/pub/automated-publishing#configuring-automated-publishing-from-github-actions-on-pub-dev

ookami-kb:
I would say, it should only run when the new version is merged into master.

witwash:
Do you know if it is supported or I need to do hacks like [this](https://samwize.com/2022/10/28/running-github-action-when-tag-pushed-on-a-specific-branch/)?

ookami-kb:
IIRC, yeah, it requires some hacks. And I'm not sure if this specific solution will work when the tag is first pushed to feature branch, and then merged into master. So let's probably go with environments instead: https://dart.dev/tools/pub/automated-publishing#hardening-security-with-github-deployment-environments

witwash:
@ookami-kb, I believe now you have to set up protection rules for the new environment. I've named it according to the example: `pub.dev`

ookami-kb (approved):

PR #640, opened 2024-07-13
feat!: Upgrade to DCM 1.19.1

witwash:
#### Summary

- upgraded to DCM 1.19.1
##### New rules: 
- [avoid-incorrect-image-opacity](https://dcm.dev/docs/rules/flutter/avoid-incorrect-image-opacity/)

##### Rules left disabled:
- [handle-throwing-invocations](https://dcm.dev/docs/rules/common/handle-throwing-invocations/) - is focused around `dart-code-metrics-annotations` package which we don't use
- [prefer-correct-throws](https://dcm.dev/docs/rules/common/prefer-correct-throws/) - again, focused around the `dart-code-metrics-annotations` package
- [prefer-extracting-function-callbacks](https://dcm.dev/docs/rules/common/prefer-extracting-function-callbacks/) - is a good idea, we could enable it (with a limit of something like 5 lines) but it is flagging a bunch of tests.

#### Testing steps

None

#### Follow-up issues

None

#### Check during review

- Verify against Jira issue.
- Unused code removed.
- Build passing.

### DIFF
--- a/.github/workflows/dcm.yml
+++ b/.github/workflows/dcm.yml
@@ -16,7 +16,7 @@ jobs:
         uses: CQLabs/setup-dcm@v1
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
-          version: "1.18.1"
+          version: "1.19.1"
 
       - uses: ./.github/actions/setup
 
--- a/mews_pedantic/lib/analysis_options.yaml
+++ b/mews_pedantic/lib/analysis_options.yaml
@@ -385,6 +385,7 @@ dart_code_metrics:
     # - format-comment
     # - format-test-name
     - function-always-returns-null
+    # - handle-throwing-invocations
     # - map-keys-ordering
     # - match-class-name-pattern
     - match-getter-setter-field-names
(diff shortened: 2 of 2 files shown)


ookami-kb:
> [handle-throwing-invocations](https://dcm.dev/docs/rules/common/handle-throwing-invocations/) - is focused around dart-code-metrics-annotations package which we don't use
[prefer-correct-throws](https://dcm.dev/docs/rules/common/prefer-correct-throws/) - again, focused around the dart-code-metrics-annotations package

@MewsSystems/tech-mo  FYI, I plan to migrate to it (I was actively asking for these rules), but it can take some time and should be done gradually. I think, we will enable them locally first for some packages, and later turn them on in `mews-pedantic`.

ookami-kb:
> [prefer-extracting-function-callbacks](https://dcm.dev/docs/rules/common/prefer-extracting-function-callbacks/) - is a good idea, we could enable it (with a limit of something like 5 lines) but it is flagging a bunch of tests.

I'm not a fan of rules like "do smth after function is N lines of code". In the majority of cases, it should rather be based on the meaning rather than its sized.

ookami-kb (approved):

chaeMil (approved):

PR #649, opened 2024-07-31
ci: Update publish action

witwash:
#### Summary

- add missing environment property

#### Testing steps

- create a release that would trigger this action. Publish should be successful.

#### Follow-up issues

- create a single action that would work depending on the input and trigger

#### Check during review

- Verify against Jira issue.
- Unused code removed.
- Build passing.


witwash:
@chaeMil ping

chaeMil (approved):

PR #658, opened 2024-08-12
feat!: Upgrade to DCM 1.21.0

witwash:
#### Summary

#### Testing steps

#### Follow-up issues

#### Check during review

- Verify against Jira issue.
- Unused code removed.
- Build passing.


ookami-kb:
> [prefer-boolean-prefixes](https://dcm.dev/docs/rules/common/prefer-boolean-prefixes/)

I think it's pretty useful and follows our convention.

> [newline-before-method](https://dcm.dev/docs/rules/common/newline-before-method/)

I would enable that.

witwash:
I was doubting about those two. I'll have to do a breaking refactor in `optimus` because some parameters are not following this exact convention

witwash:
Here is the list of the default prefixes defined in this rule: `[is, are, was, were, has, have, had, can, should, will, do, does, did]` do we want to add something? @alboiuvlad29 @chaeMil @ookami-kb

remote_logger/lib/src/remote_logger.dart
alboiuvlad29, line 1: On a side note would you be able to bump http in remote logger if you get the chance? It's quite outdated and forces us to override the version for kiosk. It also seems to keep throwing errors in development with not being able to find the domain.

witwash:
@ookami-kb,
OK, so, I've done this one with 3 commits because it is a breaking change for `optimus`, but it is not breaking for `remote_logger` and `kiosk_mode`

optimus/lib/src/dropdown/base_dropdown_tile.dart
```diff
@@ -54,7 +54,7 @@ class BaseDropdownTile extends StatelessWidget {
                     padding: EdgeInsets.only(right: tokens.spacing200),
                     child: CheckboxTick(
                       isEnabled: true,
-                      onChanged: (_) {},
+                      onChanged: (isChecked) {},
```
ookami-kb, line 57: This shouldn't be changed. `_` is unused parameter. It looks like should be handled by DCM.

witwash: Thought so

witwash:
Look like `1.21.1` is already out, but unused `_` is still flagged. Oh well 😢

optimus/lib/src/dropdown/base_dropdown_tile.dart
witwash: Fixed

optimus/lib/src/dropdown/base_dropdown_tile.dart
```diff
@@ -54,6 +54,7 @@ class BaseDropdownTile extends StatelessWidget {
                     padding: EdgeInsets.only(right: tokens.spacing200),
                     child: CheckboxTick(
                       isEnabled: true,
+                      // ignore: prefer-boolean-prefixes, DCM bug
```
ookami-kb, line 57: Can we just add `_` to `ignored-names`?

witwash: `ignored-names` is for methods
> Set ignored-names (default is empty) to ignore specific function / method names.

I've tried, no luck

ookami-kb: Ok, I've reported it back to DCM, let's have it ignored for now.

ookami-kb (approved):

PR #677, opened 2024-09-15
feat: [DX-2168] Improve password StatusBar transition animation

witwash:
#### Summary

- improved transition from color to transparency

#### Testing steps

Dev-tested

#### Follow-up issues

None

#### Check during review

- Verify against Jira issue.
- Unused code removed.
- Build passing.


witwash:
@chaeMil, can you look at this one? When you go to `Color.transparent` the animation is not fluid because the `transparent` color is derived from `black`. This change will make the transition smooth, because it would use the previous state color and add alpha, instead of the whole color -> transparent transition.

chaeMil (approved):

PR #693, opened 2024-10-03
fix: [DX-2199] Fix alignment addon

witwash:
#### Summary

- Fixed `AlignmentAddon` not working properly because of the wrong order in nesting.

#### Testing steps

- `AlignmentAddon` should work now and change align should visually change the position of the showcased component.

#### Follow-up issues

- Rework some use cases that wrap components in `Center` as they would intervene with this addon.

#### Check during review

- Verify against Jira issue.
- Unused code removed.
- Build passing.


witwash:
@chaeMil, ping

chaeMil (approved):

Issue #694, opened 2024-10-04
[Android] "Execution failed for task ':kiosk_mode:compileDebugKotlin'" ('compileDebugJavaWithJavac' (1.8))


Kuhlemann:
Android Studio Ladybug (2024.2.1) seems to be shipped with Java 21.

Since I updated Android Studio to Ladybug I get the following error when compiling my Android version:
> Inconsistent JVM-target compatibility detected for tasks 'compileDebugJavaWithJavac' (1.8) and 'compileDebugKotlin' (21).

When I remove kiosk_mode from pubspec.yaml my build (with a lot of other pub.dev-packages) compiles without any problems.
So the problem only occurs within kiosk_mode.

android\app\build.grade:
```
    compileOptions {
        sourceCompatibility = JavaVersion.VERSION_21
        targetCompatibility = JavaVersion.VERSION_21
    }

    kotlinOptions {
        jvmTarget = JavaVersion.VERSION_21
    }
```

android\gradle\gradle-wrapper.properties:
```
distributionBase=GRADLE_USER_HOME
distributionPath=wrapper/dists
distributionUrl=https\://services.gradle.org/distributions/gradle-8.10.2-all.zip
networkTimeout=10000
validateDistributionUrl=true
zipStoreBase=GRADLE_USER_HOME
zipStorePath=wrapper/dists
```

android\settings.gradle:
```
plugins {
    id "dev.flutter.flutter-plugin-loader" version "1.0.0"
    id "com.android.application" version "8.7.0" apply false
    id "org.jetbrains.kotlin.android" version "1.9.24" apply false
}
```

alboiuvlad29:
Should be fixed by https://github.com/MewsSystems/mews-flutter/pull/695.

Issue #681, opened 2024-09-16
kiosk_mode: android:attr/lStar not found when assembling in release mode


TiagoLacerda:
Whenever I try to build my flutter application, which depends on kiosk_mode ^0.5.0+5, in release mode, I get the following error:

```
Execution failed for task ':kiosk_mode:verifyReleaseResources'.
> A failure occurred while executing com.android.build.gradle.tasks.VerifyLibraryResourcesTask$Action
   > Android resource linking failed
     ERROR:<path_to_my_project>\build\kiosk_mode\intermediates\merged_res\release\values\values.xml:194: AAPT: error: resource android:attr/lStar not found.
```
     
My `project/android/build.gradle` has `ext.kotlin_version = '1.7.10'` and my `project/android/app/build.gradle` has both `compileSdkVersion flutter.compileSdkVersion` and `targetSdkVersion flutter.targetSdkVersion` for flutter 3.24.3. However, changing to the latest kotling release (2.0.20) or specifying the `targetSdkVersion` and `compileSdkVersion` to 34 have not solved this problem.

This problem does not occur when building in debug mode.

ijascookee:
Same here

alboiuvlad29:
Should be fixed by https://github.com/MewsSystems/mews-flutter/pull/695.

PR #695, opened 2024-10-07
feat: Upgrade `kiosk_mode`

alboiuvlad29:
#### Summary

Upgrades `kiosk_mode`:
- To use the latest available AGP
- To use the latest Kotlin version
- Updates the example app to use [imperative plugin apply](https://docs.flutter.dev/release/breaking-changes/flutter-gradle-plugin-apply)
- Replaces deprecated `.buildDir` with `.layout.buildDirectory`.

The changes are accurate, as they're taken from a freshly created app and plugin using `flutter create`. This should future-proof the package and allow the migrations to higher AGP versions in client flutter apps.

Fixes:
- https://github.com/MewsSystems/mews-flutter/issues/694
- https://github.com/MewsSystems/mews-flutter/issues/681

#### Testing steps

Ensure that `kiosk_mode` package and example app work as expected. I've tested the example app.

#### Follow-up issues

No.

#### Check during review

- Verify against Jira issue.
- Unused code removed.
- Build passing.

### DIFF
--- a/kiosk_mode/android/build.gradle
+++ b/kiosk_mode/android/build.gradle
@@ -32,12 +32,12 @@ android {
     compileSdk = 34
 
     compileOptions {
-        sourceCompatibility = JavaVersion.VERSION_1_8
-        targetCompatibility = JavaVersion.VERSION_1_8
+        sourceCompatibility = JavaVersion.VERSION_21
+        targetCompatibility = JavaVersion.VERSION_21
     }
 
     kotlinOptions {
-        jvmTarget = JavaVersion.VERSION_1_8
+        jvmTarget = JavaVersion.VERSION_21
     }
 
     sourceSets {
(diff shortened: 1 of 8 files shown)


Sevastyan (commented):
Let's take it a step further and finally start using a version catalog.

https://docs.gradle.org/current/userguide/dependency_management_basics.html#version_catalog

alboiuvlad29:
@Sevastyan Don't believe it's working properly with the flutter plugin. See this [diff](https://github.com/MewsSystems/mews-flutter/compare/kiosk_mode_update...kiosk_mode_update_with_version_catalogue).

image


For some reason it's expecting the same libraries declarations in the example app, while it shouldn't. It also doesn't work with `settings.gradle` as the evaluation happens after this file.

Sevastyan (approved):
Let's not this stop the PR from being merged.

However, keep in mind that version catalog is the standard we should migrate to sometime.

PR #717, opened 2024-12-09
feat!: [DX-3563] Upgrade to DCM 1.25.0

witwash:
#### Summary

Upgrade to DCM `1.25.0`

 #### Enabled rules:
Common:
[avoid-unnecessary-patterns](https://dcm.dev/docs/rules/common/avoid-unnecessary-patterns/)

Flutter:
[prefer-container](https://dcm.dev/docs/rules/flutter/prefer-container/)

Bloc:
[emit-new-bloc-state-instances](https://dcm.dev/docs/rules/bloc/emit-new-bloc-state-instances/)
[avoid-bloc-public-fields](https://dcm.dev/docs/rules/bloc/avoid-bloc-public-fields/) 
[prefer-bloc-state-suffix](https://dcm.dev/docs/rules/bloc/prefer-bloc-state-suffix/) 

[prefer-bloc-event-suffix](https://dcm.dev/docs/rules/bloc/prefer-bloc-event-suffix/) 

[avoid-instantiating-in-bloc-value-provider](https://dcm.dev/docs/rules/bloc/avoid-instantiating-in-bloc-value-provider/)
[avoid-existing-instances-in-bloc-provider](https://dcm.dev/docs/rules/bloc/avoid-existing-instances-in-bloc-provider/)

#### Rules left disabled:

[avoid-single-field-destructuring](https://dcm.dev/docs/rules/common/avoid-single-field-destructuring/) and
[prefer-assigning-await-expressions](https://dcm.dev/docs/rules/common/prefer-assigning-await-expressions/) - there are a lot of cases where it is used in `tech-mo`. I am not sure it is worth refactoring
[record-fields-ordering](https://dcm.dev/docs/rules/common/record-fields-ordering/) - I don't think it is needed
[avoid-returning-value-from-cubit-methods](https://dcm.dev/docs/rules/bloc/avoid-returning-value-from-cubit-methods/) - we have `avoid-cubits` and don't use them, so this one is not needed as well
[prefer-immutable-bloc-state](https://dcm.dev/docs/rules/bloc/prefer-immutable-bloc-state/) - doesn't support `freezed`
[prefer-immutable-bloc-events](https://dcm.dev/docs/rules/bloc/prefer-immutable-bloc-events/) - doesn't support `freezed`

#### Testing steps

#### Follow-up issues

#### Check during review

- Verify against Jira issue.
- Unused code removed.
- Build passing.

### DIFF
--- a/.github/workflows/dcm.yml
+++ b/.github/workflows/dcm.yml
@@ -16,7 +16,7 @@ jobs:
         uses: CQLabs/setup-dcm@v1
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
-          version: "1.24.2"
+          version: "1.25.0"
 
       - uses: ./.github/actions/setup
 
--- a/mews_pedantic/lib/analysis_options.yaml
+++ b/mews_pedantic/lib/analysis_options.yaml
@@ -362,6 +362,7 @@ dart_code_metrics:
     - avoid-shadowed-extension-methods
     # - avoid-shadowing
     # - avoid-similar-names
+    # - avoid-single-field-destructuring
     # - avoid-slow-collection-methods
     # - avoid-substring
     - avoid-suspicious-super-overrides
@@ -388,6 +389,7 @@ dart_code_metrics:
     - avoid-unnecessary-negations
     - avoid-unnecessary-nullable-return-type
     - avoid-unnecessary-overrides
+    - avoid-unnecessary-patterns
     - avoid-unnecessary-reassignment
     - avoid-unnecessary-return
     - avoid-unnecessary-super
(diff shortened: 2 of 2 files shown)


ookami-kb:
Do [prefer-immutable-bloc-state](https://dcm.dev/docs/rules/bloc/prefer-immutable-bloc-state/) and [prefer-immutable-bloc-events](https://dcm.dev/docs/rules/bloc/prefer-immutable-bloc-events/) work correctly with freezed classes (i.e. does it recognize them as immutable)?

witwash:
@ookami-kb,
Well, it does not 
CleanShot 2024-12-10 at 11 55 58@2x
But it can be auto-fixed or fixed from the suggestion by adding `@immutable`

ookami-kb:
@witwash I would rather disable these rules then. We always use freezed for both state and events anyway, so this seems like an extra work.

ookami-kb (approved):

PR #721, opened 2025-01-09
feat!: [DX-3716] Upgrade to DCM 1.26.1

witwash:
#### Summary

- Upgrade to DCM [1.26.0](https://dcm.dev/changelog/#1.26.0)

Enabled rules: 

- [avoid-commented-out-code](https://dcm.dev/docs/rules/common/avoid-commented-out-code/)
- [avoid-future-ignore](https://dcm.dev/docs/rules/common/avoid-future-ignore/)
- [prefer-switch-expression](https://dcm.dev/docs/rules/common/prefer-switch-expression/) 
- [avoid-continue](https://dcm.dev/docs/rules/common/avoid-continue/)
- [use-existing-destructuring](https://dcm.dev/docs/rules/common/use-existing-destructuring/)
- [use-closest-build-context](https://dcm.dev/docs/rules/flutter/use-closest-build-context/)
- [prefer-transform-over-container](https://dcm.dev/docs/rules/flutter/prefer-transform-over-container/) 
- [prefer-align-over-container](https://dcm.dev/docs/rules/flutter/prefer-align-over-container/)
- [prefer-immutable-bloc-events](https://dcm.dev/docs/rules/bloc/prefer-immutable-bloc-events) - seems to be working with `freezed` after `1.26.0`
- [prefer-immutable-bloc-state](https://dcm.dev/docs/rules/bloc/prefer-immutable-bloc-state) - seems to be working with `freezed` after `1.26.0`

Rules left disabled:
- [avoid-unnecessary-continue](https://dcm.dev/docs/rules/common/avoid-unnecessary-continue/) - used `avoid-continue` instead
- [pattern-fields-ordering](https://dcm.dev/docs/rules/common/pattern-fields-ordering/) 
- [prefer-class-destructuring](https://dcm.dev/docs/rules/common/prefer-class-destructuring/) - `use-existing-destructuring` felt more suitable, for cases when you don't want to destructure everything at once. For example, I would have to declare all tokens used for button styling and then use only those applicable to the state. Instead of just calling `tokens.textDisabled`.
- [avoid-if-with-many-branches](https://dcm.dev/docs/rules/common/avoid-if-with-many-branches/) 
- [prefer-static-method](https://dcm.dev/docs/rules/common/prefer-static-method/)
- [avoid-complex-arithmetic-expressions](https://dcm.dev/docs/rules/common/avoid-complex-arithmetic-expressions/)
- [prefer-spacing](https://dcm.dev/docs/rules/flutter/prefer-spacing/) - thinking about that one. It is useful, but sometimes spacing between widgets is not consistent.
- [prefer-commenting-future-delayed](https://dcm.dev/docs/rules/common/prefer-commenting-future-delayed/) - in most cases code is self-describing. If it is not, future delayed or not it has to be commented
- [no-empty-string](https://dcm.dev/docs/rules/common/no-empty-string/)

#### Testing steps

None

#### Follow-up issues

Release the new version

#### Check during review

- Verify against Jira issue.
- Unused code removed.
- Build passing.

### DIFF
--- a/.github/workflows/dcm.yml
+++ b/.github/workflows/dcm.yml
@@ -16,7 +16,7 @@ jobs:
         uses: CQLabs/setup-dcm@v1
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
-          version: "1.25.0"
+          version: "1.26.1"
 
       - uses: ./.github/actions/setup
 
--- a/mews_pedantic/lib/analysis_options.yaml
+++ b/mews_pedantic/lib/analysis_options.yaml
@@ -266,8 +266,11 @@ dart_code_metrics:
     - avoid-collection-equality-checks
     - avoid-collection-methods-with-unrelated-types
     # - avoid-collection-mutating-methods
+    - avoid-commented-out-code
+    # - avoid-complex-arithmetic-expressions
     - avoid-complex-loop-conditions
     - avoid-conditions-with-boolean-literals
+    # - avoid-continue
     - avoid-contradictory-expressions
     # - avoid-declaring-call-method
     - avoid-double-slash-imports
@@ -291,6 +294,7 @@ dart_code_metrics:
     # - avoid-explicit-type-declaration
     # - avoid-extensions-on-records
     - avoid-function-type-in-records
+    - avoid-future-ignore
     - avoid-future-tostring
     - avoid-generics-shadowing
     - avoid-getter-prefix:
@@ -298,6 +302,7 @@ dart_code_metrics:
     - avoid-global-state
     # - avoid-high-cyclomatic-complexity
     # - avoid-identical-exception-handling-blocks
+    # - avoid-if-with-many-branches
     # - avoid-ignoring-return-values
     - avoid-implicitly-nullable-extension-types
     # - avoid-importing-entrypoint-exports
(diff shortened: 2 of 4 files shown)


ookami-kb:
1. I would rather use `avoid-unnecessary-continue` instead of `avoid-continue`.
2. `prefer-static-method` – why not? Seems useful.
3. `prefer-spacing` – does it not take into account different spaces?

witwash:
1. sure, no problem
2. because [prefer-static-class](https://dcm.dev/docs/rules/common/prefer-static-class/) isn't enabled. I thought we didn't want static methods as well
3. I meant that `prefer-spacing` would warn about a single `SizedBox` inside a `Flex`. Using `spacing` would add the same spacing between children, which is not that common. 

I've checked it and it has some false positives, for example: 
CleanShot 2025-01-10 at 16 19 09@2x
I don't know what is triggering that, `Label` and `Error` are wrapped in `Padding` at the top of their `build`, but the warning persists even after removing `Padding`.

ookami-kb:
1. ✅ 
2. I'd say it's a different thing. Static class is rather code-style issue, here it actually shows that the method doesn't depend on class instance. It's minor thing though, I don't have a strong preference.
3. ✅

witwash:
@ookami-kb, 
`prefer-static-method` marks methods that rely on `context` or bounded `widget`s. I can rewrite most of them. Do you think it would be that beneficial? 
CleanShot 2025-01-13 at 11 38 38@2x

witwash:
@ookami-kb,
`1.26.1` is out, but the bug persists. I'll mark it for future consideration when a new patch version is out

witwash:
I've reported the bug on the DCM Discord. I'll add this rule back when the fix is up

witwash:
@ookami-kb, ready

ookami-kb (approved):

PR #728, opened 2025-01-22
feat: [DX-3888] Update ruleset

witwash:
#### Summary

- bumped DCM to `1.26.2`
- added `contains` to the list of valid prefixes
- added `prefer-static-method`
- fixed linter issues

#### Testing steps

None

#### Follow-up issues

None

#### Check during review

- Verify against Jira issue.
- Unused code removed.
- Build passing.


witwash:
@chaeMil 🥺

witwash:
Let's wait on it for a bit. DCM author will add an option to ignore private methods from the `prefer-static-method` rule 👍

PR #746, opened 2025-02-11
feat: [DX-3474] Update OptimusTabBar

witwash:
#### Summary

- padding is now set up in the `TabBar` instead for each `OptimusTab`
- the padding could be overridden
- added a possibility to enable scrolling the `OptimusTabBar`
- `OptimusTabBar` will try to fill the screen if `isScrollable` is `false`.
- updated story

#### Testing steps

1. Open `OptimusTab` story
2. Test the component using exposed knobs
3. Look for obvious visual/behavior bugs

#### Follow-up issues

None

#### Check during review

- Verify against Jira issue.
- Unused code removed.
- Build passing.

### DIFF
--- a/optimus/lib/src/tabs.dart
+++ b/optimus/lib/src/tabs.dart
@@ -35,7 +35,6 @@ class OptimusTab extends StatelessWidget {
     final tokens = context.tokens;
 
     return Container(
-      padding: EdgeInsets.symmetric(horizontal: tokens.spacing150),
       height: tokens.sizing600,
       constraints: BoxConstraints(maxWidth: maxWidth ?? double.infinity),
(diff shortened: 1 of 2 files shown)


witwash:
@chaeMil, I've noticed that you're using `Tab` in `AccountScreen` and I was wondering why is that and not `OptimusTab`?

chaeMil (approved):

chaeMil:
> @chaeMil, I've noticed that you're using `Tab` in `AccountScreen` and I was wondering why is that and not `OptimusTab`?

No idea right now, will try to add OptimusTabBar after new Optimus release and see if it fits ;)

PR #748, opened 2025-02-24
feat!: [DX-4059] Upgrade to DCM 1.27.1

witwash:
#### Summary

New rules:
[prefer-redirecting-superclass-constructor](https://dcm.dev/docs/rules/common/prefer-redirecting-superclass-constructor/)
[prefer-expect-later](https://dcm.dev/docs/rules/common/prefer-expect-later/)
[avoid-nested-try-statements](https://dcm.dev/docs/rules/common/avoid-nested-try-statements/)
[avoid-unnecessary-block](https://dcm.dev/docs/rules/common/avoid-unnecessary-block/)
[prefer-static-methods](https://dcm.dev/docs/rules/common/prefer-static-methods/) - since `1.27.0` it is possible to configure this rule to ignore private methods and we can finally enable it ;)

#### Testing steps

None

#### Follow-up issues

Publish the new package.

#### Check during review

- Verify against Jira issue.
- Unused code removed.
- Build passing.


ookami-kb:
[prefer-redirecting-superclass-constructor](https://dcm.dev/docs/rules/common/prefer-redirecting-superclass-constructor/) doesn't work with freezed classes and doesn't handle the situation when specific subclass should be passed to the function, I've reported it, so I suggest turning it off for now.

Otherwise, LGTM.

witwash:
@ookami-kb, done ;)

ookami-kb (approved):

PR #760, opened 2025-03-23
feat: [DX-6152] Updated OptimusSplitButton design

witwash:
#### Summary

- added a possibility to define a border and shape for the `DropdownButton` and `BaseButton`
- added a custom `OutlinedBorder` that would enable us to skip 0-4 borders, as it is not possible to do with the available borders https://github.com/flutter/flutter/issues/68368
- this was made so we could disable borders and make the split button look like one button with two sections instead of two buttons close to each other.

<details><summary>Preview</summary>
Before:

https://github.com/user-attachments/assets/eefb9402-e6b8-4a73-9931-8f873d6fb4df

After:

https://github.com/user-attachments/assets/1bfd89d5-d391-4cff-b27f-82d760a7597d

</details>

#### Testing steps
1. Open OptimusSplitButton use case
2. Test the component using exposed knobs
#### Follow-up issues

#### Check during review

- Verify against Jira issue.
- Unused code removed.
- Build passing.

### DIFF
--- a/optimus/lib/src/button/base_button.dart
+++ b/optimus/lib/src/button/base_button.dart
@@ -1,7 +1,11 @@
+import 'package:dfunc/dfunc.dart';
 import 'package:flutter/material.dart';
 import 'package:optimus/optimus.dart';
 import 'package:optimus/src/button/base_button_variant.dart';
 import 'package:optimus/src/button/common.dart';
+
+typedef ShapeBuilder =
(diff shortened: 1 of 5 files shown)


witwash:
@chaeMil, ping

witwash:
@chaeMil, what's up with this one? I thought we'd agreed I would fix the overflow you found in the other ticket.

chaeMil (approved):

PR #818, opened 2025-07-07
feat: [DX-1645] Add text scaling to dropdown menus

witwash:
#### Summary

- updated dropdowns to respect custom `textScale`

The scaling to 2x is an extreme that we have to support due to accessibility guidelines. It does not have to be pretty, but it has to be functional and not block users from proceeding further.

#### Testing steps

1. Set up custom text scale (Addon panel, `TextScale` slider)
2. Open the component that has a dropdown (for example DropdownButton or Search)
3. Test it using exposed knobs.

#### Follow-up issues

None

#### Check during review

- Verify against Jira issue.
- Unused code removed.
- Build passing.


witwash:
This one is waiting for #809, then I'll complete te rework.

chaeMil (approved):

PR #839, opened 2025-09-02
feat!: [DX-7827] Upgrade to DCM 1.31.4

witwash:
##### New rules:
Common:
[avoid-unnecessary-local-variable](https://dcm.dev/docs/rules/common/avoid-unnecessary-local-variable/)
[avoid-unnecessary-late-fields](https://dcm.dev/docs/rules/common/avoid-unnecessary-late-fields/) 
[avoid-unnecessary-nullable-parameters](https://dcm.dev/docs/rules/common/avoid-unnecessary-nullable-parameters/) 
[avoid-never-passed-parameters](https://dcm.dev/docs/rules/common/avoid-never-passed-parameters/)
[prefer-returning-condition](https://dcm.dev/docs/rules/common/prefer-returning-condition/) 
[avoid-wildcard-cases-with-sealed-classes](https://dcm.dev/docs/rules/common/avoid-wildcard-cases-with-sealed-classes/)

And [avoid-default-tostring](https://dcm.dev/docs/rules/common/avoid-default-tostring/), since they added a possibility to ignore enums.

Flutter:
[prefer-constrained-box-over-container](https://dcm.dev/docs/rules/flutter/prefer-constrained-box-over-container/) 
[prefer-void-callback](https://dcm.dev/docs/rules/flutter/prefer-void-callback/) 
[prefer-async-callback](https://dcm.dev/docs/rules/flutter/prefer-async-callback/) 

#### Testing steps

None

#### Follow-up issues

Release

#### Check during review

- Verify against Jira issue.
- Unused code removed.
- Build passing.

### DIFF
--- a/.github/workflows/dcm.yml
+++ b/.github/workflows/dcm.yml
@@ -16,7 +16,7 @@ jobs:
         uses: CQLabs/setup-dcm@v1
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
-          version: "1.30.1"
+          version: "1.31.4"
 
       - uses: ./.github/actions/setup
 
--- a/mews_pedantic/lib/analysis_options.yaml
+++ b/mews_pedantic/lib/analysis_options.yaml
@@ -280,7 +280,8 @@ dart_code_metrics:
     # - avoid-continue
     - avoid-contradictory-expressions
     # - avoid-declaring-call-method
-    # - avoid-default-tostring
+    - avoid-default-tostring:
+        ignore-enums: true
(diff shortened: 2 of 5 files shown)


witwash:
As always, eventually you'll have to update to this version and this will influence your projects. So feedback is welcomed 👂

alboiuvlad29 (approved):
When is "eventually"? 😭

witwash:
I arranged some time for the migration with my manager. I'm going to do it in this Q

